PR #3, opened 2026-02-26
docs: symlink root README to crate README and add book link

jlizen:
[visualized](https://github.com/jlizen/battery-pack/tree/update-readme)


jlizen:
cc @nikomatsakis  (don't see a way to tag PR workflow)

PR #6, opened 2026-02-28
fix: fix a lot of clippy lints

JayanAXHF:
Just some cleanup. In some places i wasn't sure so i allowed the lints. Feel free to change that


jlizen (approved):
Thanks for this... I've been averting my eyes every build

PR #7, opened 2026-02-28
ops: add some rudimentary CI to prevent lints and such in the repo

JayanAXHF:
Pretty self explanatory again, this would be nice to have during when this project is growing. We can always change to a better approach when the app becomes stable

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -16,7 +16,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Build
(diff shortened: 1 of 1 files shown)


jlizen:
Ah yeah, looks like we need either typos fixed or an allow dictionary. Glad to merge once it is green (or feel free if you have write perms)

JayanAXHF:
I've added a `_typos.toml`, so it should be fine now. I don't have write perms yet

JayanAXHF:
Oh thank god there are no more typos 🙏

PR #49, opened 2026-03-10
feat(template): Allow `bp-managed = true` in template dependencies

### LINKED ISSUE #34: [Templates] Don't require the template to specify bp-managed versions
Today, if you want to generate a template, you sort of simulate the results of `cargo bp add`, where you add both a dependency and then the bp metadata block.

Ideally, we could instead only specify the bp metadata, and on template consumption time, materialize the dependency versions based on the battery pack's own dependencies. That way we can keep a single source of truth around target version, without letting templates go stale.

You can see that in practice here: https://github.com/battery-pack-rs/battery-pack/pull/33

jlizen:
# Allow `bp-managed = true` in template dependencies

Closes #34. Stacked on #45.

### Summary

Template Cargo.toml files can now use `bp-managed = true` instead of hardcoding versions. 

This lets templates avoid hardcoding dependency versions/features, and instead defer to whichever battery pack manages them. Less drift-prone / annoying to manage. It also allows better CI validations. Consumers use `cargo bp sync` to always point to the newest battery pack dependency state, so templates should be tested against newest too.

The behavior for multiple battery packs naming a dependency is unspecified.  Technically, last write wins currently. We can add a nicer API later. I wanted to start simple.

Templates can still specify unmanaged dependencies with explicit versions/features. Those are left untouched.

Template API: 

```toml
[dependencies]
clap = { bp-managed = true } # bp-managed disallows any other keys (for now)
anyhow = "1" # unmanaged

[build-dependencies]
cli-battery-pack = { bp-managed = true }

[package.metadata.battery-pack]
cli-battery-pack = { features = ["default"] }
```

The above becomes:

```toml
[dependencies]
clap = { version = "4", features = ["derive"] }
anyhow = "1"

[build-dependencies]
cli-battery-pack = "0.4.0"

[package.metadata.battery-pack]
cli-battery-pack = { features = ["default"] }
```

At generation time (both `cargo bp new` and `cargo bp new --preview`), `bp-managed` entries are resolved from the battery pack's spec. For `cargo bp new cli-battery-pack`, that's the latest published version from crates.io. For `--path`, it's the local source tree. If a template references a battery pack that isn't in the local workspace (e.g. `cli-battery-pack` referencing `logging-battery-pack` after being downloaded from crates.io), the spec is fetched from the registry.

**Design tradeoffs:**

- Goes against "bps are just crates" ethos since we now have a special cargo flag. But, this is only when *authoring* a template, and in any rendered view (including preview), it is resolved into normal dependency declarations. We also have jinja-friendly placeholder tags that aren't native Cargo either. I think it's probably a decent trade, since it solves a real maintenance problems for bp template authors. We can iterate on the API.

- `bp-managed = true` replaces the *entire* dep entry (version + features). No other keys are allowed alongside it; `version`, `features`, `default-features`, etc. all produce an error. We considered allowing partial overrides (e.g. `version = "bp-managed"` with custom features), but that puts magic strings in fields that normally take semver values. Any tool parsing the Cargo.toml would see an invalid version. A separate boolean key is unambiguous. Explicit deps are the escape hatch. If partial overrides turn out to be needed, we can add them later.

### Testing

- Snapshot tests (expect-test) for resolved Cargo.toml output through both the `resolve_bp_managed` and `preview()` paths
- Validation tests: errors on `version` + bp-managed, `features` + bp-managed, `default-features` + bp-managed, unresolvable dep
- Noop test: explicit versions left untouched, no-managed-deps is a passthrough
- Discovery tests: workspace case (member discovers siblings) and standalone case (no workspace)
- `with_template_two_level_generation` integration test passes end-to-end
- `cargo bp validate --path battery-packs/cli-battery-pack` validates both templates

### DIFF
--- a/src/battery-pack/bphelper-cli/src/lib.rs
+++ b/src/battery-pack/bphelper-cli/src/lib.rs
@@ -1780,6 +1780,150 @@ fn resolve_local_template(local_path: &str, template: Option<&str>) -> Result<(P
     Ok((local_path.to_path_buf(), template_path))
 }
 
+// ============================================================================
+// bp-managed dependency resolution
+// ============================================================================
+
+/// Resolve `bp-managed = true` dependencies in a Cargo.toml string,
+/// returning the rewritten content with concrete versions.
+pub fn resolve_bp_managed_content(content: &str, bp_crate_root: &Path) -> Result<String> {
+    let mut doc: toml_edit::DocumentMut = content.parse().context("failed to parse Cargo.toml")?;
+
+    // Collect all bp-managed dep names across all sections.
+    let sections = ["dependencies", "dev-dependencies", "build-dependencies"];
+    let mut has_managed = false;
+    for section in &sections {
+        if let Some(table) = doc.get(section).and_then(|v| v.as_table()) {
+            for (name, value) in table.iter() {
+                if is_bp_managed(value) {
+                    has_managed = true;
+                    let extra = extra_keys_on_bp_managed(value);
+                    if !extra.is_empty() {
+                        bail!(
+                            "dependency '{}' in [{}] has `bp-managed = true` with conflicting keys: {}",
+                            name,
+                            section,
+                            extra.join(", ")
+                        );
+                    }
+                }
+            }
+        }
+    }
+
+    if !has_managed {
+        return Ok(content.to_string());
+    }
+
+    // Read active features for each battery pack from the generated manifest's metadata.
+    let raw: toml::Value = toml::from_str(content).context("failed to parse Cargo.toml")?;
+
+    // Discover battery pack specs reachable from bp_crate_root.
+    let all_specs = bphelper_manifest::discover_from_crate_root(bp_crate_root)?;
+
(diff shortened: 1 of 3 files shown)


src/battery-pack/bphelper-cli/src/lib.rs
```diff
@@ -1773,6 +1780,180 @@ fn resolve_local_template(local_path: &str, template: Option<&str>) -> Result<(P
...
+
+/// Resolve `bp-managed = true` dependencies in a generated project's Cargo.toml.
+///
+/// Walks `project_dir` to find every `Cargo.toml` and resolves any
+/// `bp-managed = true` dependencies found in each one.
+///
+/// `bp_crate_root` is the directory of the battery pack whose template was used
+/// to generate the project. It (and sibling battery packs in the same workspace)
+/// are used to resolve specs.
+pub fn resolve_bp_managed(project_dir: &Path, bp_crate_root: &Path) -> Result<()> {
```
jlizen: will fix this pub visibility in #45

(right now a bunch of things are in integ tests without needing to be)

PR #63, opened 2026-03-15
fix(lint): fix clippy unused variable

JayanAXHF:
cc: @jlizen 

This might be a bug in the render_template_dir function that the variables weren't used. I'm not sure


jlizen (approved):
Yeah, I think I swept this one up in a different PR, but feel free to merge if it persists after you rebase against main.

Issue #54, opened 2026-03-12
(refactor) Update tui templates to use a pager


jlizen:
#45 adds a quick and dirty "preview" of templates that materializes them in the tui. But, it uses crude scrolling as part of the main tui.

A pager would be better to allow searching and other nicer viewing features, as these templates might be large.

PR #66, opened 2026-04-02
feat: add page and jump scrolling to preview

shamnad-sherief:


src/battery-pack/bphelper-cli/src/tui.rs
```diff
@@ -2058,7 +2062,7 @@ fn render_preview(frame: &mut Frame, state: &PreviewScreen) {
     frame.render_widget(preview, main);
 
     frame.render_widget(
-        Paragraph::new("↑↓/jk Scroll | Esc Back").style(Style::default().white().on_dark_gray()),
+        Paragraph::new("↑↓/jk/PgUp/PgDn Scroll | Esc Back").style(Style::default().white().on_dark_gray()),
```
jlizen: No home/end just to avoid instruction overload, I assume? That's fine by me, it is fairly intuitive and eventually we can offload this to user's own pager (#54 )

jlizen (approved):
Thanks for the contribution!

PR #67, opened 2026-04-02
feat(cli): Expand capabilities

epage:


jlizen (commented):
(Haven't reviewed for content yet, seems reasonable, but will look closer once CI is green)

> cargo test did not work for me locally. I was receiving prompts.

Just fixed this in #70  if you merge main.

Looks like CI is red, previously it was failing due to something being wrong with the template's manifest, though now I see some rustc failures

jlizen (approved):
Looks great, thanks! TIL about a bunch of these crates (which I guess is the whole point)

Will cut a release shortly

PR #80, opened 2026-04-07
refactor test: change tests to use snapbox instead of expect-test

### LINKED ISSUE #52: (refactor) We should probably use insta for snapshot TUI testing
We should probably use `insta` to do snapshot testing for the TUI (similar to ui tests in r-l/r). It's quite a bit less error prone that comparing values manually.

JayanAXHF:
closes #52 

cc: @epage @jlizen


src/battery-pack/bphelper-cli/src/commands/tests.rs
```diff
@@ -1060,12 +1058,8 @@ fn add_default_crates_basic() {
...
     let deps = extract_section(&content, "[dependencies]");
 
-    expect![[r#"
-        [dependencies]
-        anyhow = "1"
-        thiserror = "2"
-    "#]]
-    .assert_eq(&deps);
+    assert!(deps.contains("anyhow = \"1\""), "Expected anyhow");
+    assert!(deps.contains("thiserror = \"2\""), "Expected thiserror");
```
epage: How come you moved away from snapshots?

src/battery-pack/bphelper-cli/src/template_engine/tests.rs
```diff
@@ -0,0 +1,299 @@
...
+#[test]
+fn parse_config_placeholder_defaults() {
+    let toml = r#"
+            [placeholders.name]
+        "#;
+    let config: BpTemplateConfig = toml::from_str(toml).unwrap();
+    let p = &config.placeholders["name"];
+    assert!(p.prompt.is_none());
+    assert!(p.default.is_none());
+    assert_eq!(p.placeholder_type, PlaceholderType::String);
```
epage, line 59: What I do in a situations like this is:
```rust
assert_data_eq!(p.to_debug(), str![]);
```

src/battery-pack/bphelper-cli/src/tui/tests.rs
```diff
@@ -0,0 +1,1593 @@
...
+                make_entry("sqlx", false, true), // changed
+            ],
+        ),
+    ]);
+
+    let changes = state.collect_changes();
+    let actual = format!("{:#?}", changes);
+    assert_data_eq!(
+        actual,
+        snapbox::file!["snapshots/collect_changes_skips_unchanged_packs.snap"]
```
epage: `actual.to_debug()` so you don't have to do `format!("{:#?}", changes)`

src/battery-pack/bphelper-cli/src/commands/tests.rs
JayanAXHF: This was much less flackey than using snapshots for this kind of thing in my experience

epage: For cargo, when we switched to `snapbox`, we got rid of almost all of our `contains` checks because we found that they were a common source of manual updates and it wasn't an issue to cover too much output in our tests because it was just a snapshot-update away from being fixed.

jlizen: ++ i tend to prefer snapshots, it makes it a lot easier to visualize how things are changing in future PRs

src/battery-pack/bphelper-cli/src/commands/tests.rs
```diff
@@ -1143,14 +1133,13 @@ fn add_with_named_feature_records_metadata() {
...
 
-    expect![[r#"
-        [package.metadata.battery-pack.fancy-battery-pack]
-        features = [
-            "default",
-            "indicators",
-        ]
-    "#]]
-    .assert_eq(&meta);
+    assert!(
```
jlizen: We could use `str![]` if preferred instead of standalone file snapshots. Either would be stronger than `contains` if we do the bit of legwork to mask out the dynamic bits.

jlizen (commented):
Agree with ed's feedback around using snapshots, and some of the debug incantations.

src/battery-pack/bphelper-cli/src/commands/tests.rs
JayanAXHF: Indeed that makes sense, im changing these contains tests to use a snapshot now

src/battery-pack/bphelper-cli/src/template_engine/tests.rs
JayanAXHF: Done, thanks

src/battery-pack/bphelper-cli/src/commands/tests.rs
JayanAXHF: The generated file snapshot contains 
```txt
[package.metadata.battery-pack.fancy-battery-pack]
features = [
    "default",
    "indicators",
]
```
which to me seems reasonable to include as-is

src/battery-pack/bphelper-cli/src/tui/tests.rs
JayanAXHF: Done!

src/battery-pack/bphelper-cli/src/commands/tests.rs
jlizen: Cool, re-request me when you're ready for a final stamp, going AFK :)

src/battery-pack/bphelper-cli/src/commands/tests.rs
JayanAXHF: Alright, i've added this

JayanAXHF:
There are still tests that need to be migrated from contains ( a lot of them), but i think that can be done incrementally over multiple PRs

jlizen (approved):
> There are still tests that need to be migrated from contains ( a lot of them), but i think that can be done incrementally over multiple PRs

Seems fine, mind opening a backlog issue though?

JayanAXHF:
Created #83 @jlizen

PR #81, opened 2026-04-10
chore(deps): upgrade ratatui to 0.30 and enable snapbox term-svg

orhun:
fixes https://github.com/battery-pack-rs/battery-pack/pull/67/changes#r3029920901

these should probably be separate changes (commits) but they depend on each other


jlizen (approved):
Appreciate it!

PR #84, opened 2026-04-14
feat: add cargo-bp binary crate for cargo install

jlizen:
### Summary

`cargo install battery-pack` required `--features cli` because the binary was gated behind an optional feature on the library crate. Cargo has no way to distinguish "installing as a binary" from "adding as a dependency," so the only clean fix is a separate crate.

`cargo-bp` is a thin wrapper (one-line `main.rs`) that depends on `bphelper-cli` directly. `cargo install cargo-bp` now just works. The `battery-pack` library crate is unchanged, so there is no breaking change for existing consumers.

### Testing

Moved the `new_suffix` and `with_template` integration tests from the library to the CLI crate since they exercise the `cargo-bp` binary. Updated path helpers to resolve the library crate's template directory from the new location.


JayanAXHF (commented):
Have you removed the cli feature on battery-pack then? I can't tell as the diffs are horrendous on mobile

jlizen:
No, it is still there, since the validate helpers use it. If we were to split that out, we could remove the feature, but I didn't do it here.

nikomatsakis (approved):

PR #88, opened 2026-04-17
Fix detail view not scrolling when selection moves off screen

nikomatsakis:
render_selectable_section now returns the line index of the selected item. render_detail uses this to compute a scroll offset so the Paragraph keeps the selection visible.


jlizen (approved):

jlizen:
I didn't test myself, but the change seems reasonable, approving under the assumption you did

nikomatsakis:
I did.

Issue #83, opened 2026-04-11
(backlog) Refactor the tests using `contains` to use snapshots instead


JayanAXHF:
The tests that use the `.contains()` should be changed to use either a file snapshot or a `str![]` snapshot for maintainibility and less fragility.

See [`snapbox`'s docs](https://docs.rs/snapbox) for more info

Initial PR:
- #80

# Implementation History:

PR #91, opened 2026-04-18
feat(cli): add template preview to `cargo bp show -t`

jlizen:
### Summary

The `--preview` flag on `cargo bp new` was specced and documented but lost during the lib.rs module split refactor (eed5c9a). Rather than restoring it on `new`, this moves template preview to `cargo bp show -t <template>`, which is a better fit: `show` already means "show me info about this pack," and a template name just narrows what you're showing.

```bash
# Interactive TUI preview (syntax-highlighted, scrollable)
cargo bp show cli -t default

# Non-interactive (stdout, for CI or piping)
cargo bp show cli -t default --non-interactive

# With a local path
cargo bp show fancy -t default --non-interactive --path tests/fixtures/fancy-battery-pack
```

Without `-t`, `show` behaves exactly as before. The TUI's existing preview (pressing `p` on a template in the detail screen) also continues to work.

Threw in some light consolidation of related helpers.

### Testing

- e2e CLI snapshot testing for non-interactive paths
- unit tests on the CLI handling, TUI rendering, tui escape

### DIFF
--- a/md/spec/cli.md
+++ b/md/spec/cli.md
@@ -172,13 +172,6 @@ r[cli.new.define-flag]
 named placeholder to the given value, skipping the prompt for that
 placeholder. Multiple `-d` flags MAY be provided.
 
-r[cli.new.preview]
-`cargo bp new <pack> --preview` MUST render the template and print
-the resulting files to stdout without writing anything to disk.
-Placeholders without a default MUST fall back to `<name>` so the
-preview always succeeds. If `--name` is not provided, the preview
-MUST use `my-project` as the project name.
-
 r[cli.new.non-interactive]
 In non-interactive mode, `cargo bp new` MUST fail with an error
 if `--name` is not provided. Template placeholders without a
@@ -336,3 +329,12 @@ in the interactive TUI.
 r[cli.show.non-interactive]
 In non-interactive mode, `cargo bp show` MUST print results as
 plain text.
+
+r[cli.show.template-preview]
+`cargo bp show <pack> --template <name>` MUST render the named
+template and display the resulting files. In a TTY, the output
+SHOULD be shown in the interactive TUI preview screen. With
+`--non-interactive`, the rendered files MUST be printed to stdout.
+Placeholders without a default MUST fall back to `<name>` so the
+preview always succeeds. The project name MUST default to
+`my-project`.
--- a/src/battery-pack/bphelper-cli/src/commands/mod.rs
+++ b/src/battery-pack/bphelper-cli/src/commands/mod.rs
@@ -16,10 +16,9 @@ use crate::manifest::{
     write_bp_features_to_doc, write_deps_by_kind, write_workspace_refs_by_kind,
 };
 use crate::registry::{
-    CrateSource, InstalledPack, TemplateConfig, download_and_extract_crate,
-    fetch_battery_pack_detail, fetch_battery_pack_detail_from_source, fetch_battery_pack_list,
-    fetch_bp_spec, find_local_battery_pack_dir, load_installed_bp_spec, lookup_crate,
-    resolve_crate_name, short_name,
+    CrateSource, InstalledPack, TemplateConfig, fetch_battery_pack_detail,
+    fetch_battery_pack_detail_from_source, fetch_battery_pack_list, fetch_bp_spec,
+    load_installed_bp_spec, resolve_crate_name, short_name,
 };
 
 // [impl cli.bare.help]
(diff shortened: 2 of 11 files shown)


src/battery-pack/bphelper-cli/src/tui.rs
```diff
@@ -24,15 +24,61 @@ use std::time::Duration;
...
+pub(crate) struct ShowOpts<'a> {
+    pub battery_pack: &'a str,
+    pub template: Option<&'a str>,
+    pub path: Option<&'a str>,
+    pub source: CrateSource,
+}
+
 /// Run the TUI starting from the list view
-pub fn run_list(source: CrateSource, filter: Option<String>) -> Result<()> {
+pub(crate) fn run_list(source: CrateSource, filter: Option<String>) -> Result<()> {
```
jlizen, line 36: These are already in a crate-local module, so just changing the signature to be less misleading

src/battery-pack/bphelper-cli/src/tui.rs
```diff
@@ -24,15 +24,61 @@ use std::time::Duration;
...
     let app = App::new_list(source, filter);
     app.run()
 }
 
-/// Run the TUI starting from the detail view
-pub fn run_show(name: &str, path: Option<&str>, source: CrateSource) -> Result<()> {
-    let app = App::new_show(name, path, source);
+/// Run the TUI for a battery pack. Without `template`, shows the detail
+/// screen. With `template`, jumps directly to the template preview.
+pub(crate) fn run_show(opts: ShowOpts<'_>) -> Result<()> {
```
jlizen, line 43: same

md/spec/cli.md
```diff
@@ -156,13 +156,6 @@ r[cli.new.define-flag]
 named placeholder to the given value, skipping the prompt for that
 placeholder. Multiple `-d` flags MAY be provided.
```
nikomatsakis, line 173: coulda sworn I killed these files. I guess I didn't open that PR

src/cargo-bp/tests/show_preview.rs
```diff
@@ -0,0 +1,80 @@
...
+    );
+    assert!(
+        stdout.contains("── src/main.rs ──"),
+        "expected src/main.rs header in:\n{stdout}"
+    );
+    // Preview should use "my-project" as the default project name
+    assert!(
+        stdout.contains("my-project"),
+        "expected 'my-project' placeholder in:\n{stdout}"
+    );
```
nikomatsakis, line 68: I think we should include `expect_test`-based snapshots in these tests (and likely others!) in addition to the asserts above.

src/battery-pack/bphelper-cli/src/tui.rs
```diff
@@ -24,15 +24,61 @@ use std::time::Duration;
...
+    if opts.template.is_some() {
+        run_preview(opts)
+    } else {
+        let app = App::new_show(opts.battery_pack, opts.path, opts.source);
+        app.run()
+    }
+}
+
+/// Run the TUI starting directly in the template preview screen.
+fn run_preview(opts: ShowOpts<'_>) -> Result<()> {
```
nikomatsakis, line 53: Do we have a way to test this behavior?

src/battery-pack/bphelper-cli/src/commands/mod.rs
```diff
@@ -264,18 +268,37 @@ pub fn main() -> Result<()> {
...
+                        if non_interactive || !interactive {
+                            // [impl cli.show.template-preview]
+                            print_template_preview(&crate::template_engine::PreviewOpts {
+                                battery_pack: show_opts.battery_pack,
+                                template: tmpl,
+                                path: show_opts.path,
+                                source: &show_opts.source,
+                            })
+                        } else {
+                            crate::tui::run_show(show_opts)
```
nikomatsakis: It's...fine but there's something funny about the way this code works, i.e., we test here "if there is a template" and then we test *again* in `run_show`. I kind of expect something like:

```rust
if non_interactive || !interactive {
    crate::noninteractive::run_show(show_opts)
} else {
    crate::tui::run_show(show_opts)
}
```

and then each of those helpers would check, or something like that.

nikomatsakis (changes requested):
The `expect_test` in the test would be nice, the rest are thoughts and suggestions.

src/battery-pack/bphelper-cli/src/tui.rs
jlizen: I had some tests of the inner screen, but I can throw in the snapshots like you suggested

src/cargo-bp/tests/show_preview.rs
jlizen: i'll take at least a first pass on #83 separately, but fix the new ones here

nikomatsakis (approved):

nikomatsakis:
approved, but needs rebase @jlizen

PR #100, opened 2026-04-19
refactor(test): add contains assertions alongside snapshots, inline small file snapshots

jlizen:
[As discussed in Zulip](https://rust-lang.zulipchat.com/#narrow/channel/575623-battery-packs-rs/topic/Testing.20conventions.20battery-pack-rs.2Fbattery-pack.2397/with/586332083)

Adding back in the simpler "contains" assertions that the spirit of the snaps don't accidentally drift. We still keep the snaps.

I also moved many of the little snaps to be inline and managed with `str![]`. More readable to have it next to the test.


JayanAXHF:
IMO all of these `assert_eq!`s should be done with `str!`. Prob a bit more overhead but it makes updating them _much easier_

src/battery-pack/bphelper-build/src/tests.rs
```diff
@@ -211,15 +218,43 @@ A great battery pack.
 fn test_render_crate_table() {
     let ctx = simple_context();
     let output = render_docs("{{crate-table}}", &ctx).unwrap();
```
JayanAXHF, line 220: We are already checking for the table in the data_eq check, what purpose does the assert serve here?

src/battery-pack/bphelper-build/src/tests.rs
```diff
@@ -211,15 +218,43 @@ A great battery pack.
...
+
+"#]]
+    )
 }
 
 #[test]
 // [verify docgen.template.default]
 fn test_render_default_template() {
     let ctx = simple_context();
     let output = render_docs("{{readme}}\n\n{{crate-table}}", &ctx).unwrap();
```
JayanAXHF, line 241: Similar concern here, the `assert` is redundant

src/battery-pack/bphelper-build/src/tests.rs
```diff
@@ -237,7 +272,25 @@ fn test_render_custom_template() {
...
+
+A test pack
+
+## Crates
+
+- **anyhow** (1): Flexible concrete Error type
+- **thiserror** (2): derive(Error)
+
+"#]]
+    )
```
JayanAXHF, line 293:
```suggestion
    assert_data_eq!(
        output,
        str![[r#"
# test-battery-pack v0.1.0

A test pack

## Crates

- **anyhow** (1): Flexible concrete Error type
- **thiserror** (2): derive(Error)

"#]]
    )
```

src/battery-pack/bphelper-build/src/tests.rs
JayanAXHF, line 1: To me, many of these asserts look redundant: we are checking for what they check directly below it. It just adds unnecessary churn for updating imho

src/battery-pack/bphelper-cli/src/commands/tests.rs
JayanAXHF, line 1: similar concern here

JayanAXHF (changes requested):
I'm not sure; the asserts seem largely verbose and are checked for directly below it. This was one of my motives with the snapbox refactor, to make updating tests in case of updates easier. Curious to hear from @nikomatsakis

src/battery-pack/bphelper-build/src/tests.rs
jlizen: > unnecessary churn for updating imho

If there are contains that are asserting things that will change, please call 'em out, I'd like to make them lossier.

The main point here is to allow large layout shifts, with big diffs, but still pin down the "core" thing that needs to exist in the screen no matter how much it shifts.

But anyway I am open to scaling it back, interested in @nikomatsakis 's perspective too.

src/battery-pack/bphelper-build/src/tests.rs
```diff
@@ -99,6 +99,13 @@ fn test_context_hidden_excluded() {
     let ctx = build_context(&spec, &descriptions, "");
 
     let names: Vec<&str> = ctx.crates.iter().map(|c| c.name.as_str()).collect();
+    // serde, serde_json, cc should be hidden
+    assert!(!names.contains(&"serde"), "serde should be hidden");
+    assert!(
+        !names.contains(&"serde_json"),
+        "serde_json should be hidden"
+    );
+    assert!(!names.contains(&"cc"), "cc should be hidden");
```
nikomatsakis, line 108: I like seeing this. If I was reviewing a diff and I just saw `serde` being added to this list..

```rust
assert_data_eq!(
        names.join(", "),
        str!["assert_cmd, clap, console, dialoguer, indicatif, predicates, serde"]
    );
```

...it wouldn't obviously be a red flag for me. But I like having the output when I'm reading tests, it lets me put things in context.

src/battery-pack/bphelper-build/src/tests.rs
```diff
@@ -211,15 +218,43 @@ A great battery pack.
 fn test_render_crate_table() {
     let ctx = simple_context();
     let output = render_docs("{{crate-table}}", &ctx).unwrap();
-    assert_data_eq!(output, file![_])
+    assert!(
+        output.contains("| Crate |"),
+        "should render a markdown table"
+    );
```
nikomatsakis, line 224:
```suggestion
```

yeah, this particular one doesn't seem that valuable

src/battery-pack/bphelper-build/src/tests.rs
```diff
@@ -211,15 +218,43 @@ A great battery pack.
...
 }
 
 #[test]
 // [verify docgen.template.default]
 fn test_render_default_template() {
     let ctx = simple_context();
     let output = render_docs("{{readme}}\n\n{{crate-table}}", &ctx).unwrap();
-    assert_data_eq!(output, file![_])
+    assert!(output.contains("# My Pack"), "should contain readme");
+    assert!(output.contains("| Crate |"), "should contain crate table");
```
nikomatsakis, line 243:
```suggestion
```

I'm inclined to agree that this doesn't seem important enough. I'm more intereted in asserts that describe subtle properties of the output, not that the output exists at all.

src/battery-pack/bphelper-build/src/tests.rs
```diff
@@ -262,7 +315,18 @@ fn test_render_crate_table_empty() {
 fn test_render_crate_table_links() {
     let ctx = simple_context();
     let output = render_docs("{{crate-table}}", &ctx).unwrap();
-    assert_data_eq!(output, file![_]);
+    assert!(output.contains("[anyhow](https://crates.io/crates/anyhow)"));
+    assert!(output.contains("[thiserror](https://crates.io/crates/thiserror)"));
```
nikomatsakis, line 319: so like this seems good

src/battery-pack/bphelper-build/src/tests.rs
```diff
@@ -272,6 +336,12 @@ fn test_render_no_html_escaping() {
         ..simple_context()
     };
     let output = render_docs("{{readme}}", &ctx).unwrap();
+    // Angle brackets and ampersand should NOT be escaped
+    assert!(!output.contains("&amp;"), "ampersand should not be escaped");
```
nikomatsakis, line 340: and this

src/battery-pack/bphelper-build/src/tests.rs
```diff
@@ -288,7 +358,30 @@ fn test_full_pipeline_basic() {
...
-    assert_data_eq!(output, file![_]);
+    assert!(
+        output.contains("# basic-battery-pack"),
+        "should contain readme"
+    );
+    assert!(output.contains("[anyhow]"), "should contain crate links");
+    assert!(
+        output.contains("[eyre]"),
+        "should contain optional crate eyre"
+    );
```
nikomatsakis, line 369: for some reason I like this one, I guess because it's checking that the `{readme}` and `{crate-table}` are both respected

src/battery-pack/bphelper-build/src/tests.rs
```diff
@@ -324,7 +447,20 @@ fn test_crate_table_update_is_automatic() {
     // test_render_crate_table).
     let ctx = simple_context();
     let output = render_docs("{{crate-table}}", &ctx).unwrap();
-    assert_data_eq!(output, file![_]);
+    assert!(
+        output.contains("| Crate |"),
+        "crate-table helper must be registered"
+    );
```
nikomatsakis, line 441:
```suggestion
```

this doesn't seem important

src/battery-pack/bphelper-build/src/tests.rs
```diff
@@ -380,7 +542,27 @@ fn test_generate_docs_writes_output_file() {
...
     let content = std::fs::read_to_string(&output_path).unwrap();
-    assert_data_eq!(content, file![_]);
+    assert!(
+        content.contains("# Hello"),
+        "output must contain rendered readme"
+    );
+    assert!(
+        content.contains("[anyhow]"),
+        "output must contain crate table"
+    );
```
nikomatsakis, line 540:
```suggestion
```

I know I said the opposite earlier, so maybe that's invalid too, but it seems like this is duplicated from other tests and not that interesting.

src/battery-pack/bphelper-cli/src/commands/tests.rs
```diff
@@ -1439,7 +1627,20 @@ fn add_twice_is_idempotent() {
...
+    assert!(build_rs.contains("fn main()"), "Expected fn main()");
+    assert!(
+        build_rs.contains("basic_battery_pack::validate"),
+        "Expected basic_battery_pack::validate call"
+    );
+    assert_data_eq!(
+        build_rs,
+        str![[r#"
+fn main() {
+    basic_battery_pack::validate();
```
nikomatsakis: hmm, I thought we removed this? maybe not on this branch?

nikomatsakis (approved):

nikomatsakis (approved):

nikomatsakis:
@jlizen made a few tweaks but needs a rebase now

jlizen:
@nikomatsakis can you take a look at this thread? https://github.com/battery-pack-rs/battery-pack/pull/100#discussion_r3106264503

src/battery-pack/bphelper-build/src/tests.rs
JayanAXHF: Having records of even tiny layout shifts, barring the spurious ones, can still be important for keeping track of when layout changes were made, which can help us find bugs later on. And if large layouts do change, the test will still be failing regardless, so I don’t find much value in these, as the tests would need to be blessed anyhow

jlizen: Maybe let's throw them in, and then the first time one needs to be added ALONG with updating a snap, we rip that one out? Does that seem like a decent compromise?

Agreed that we want the layout guards too (I didn't get rid of any right?). Niko just also likes having the "base assertion" without layout, which I vaguely agree with. And theoretically they should not need to be kept up to date in the same way.

JayanAXHF: > Maybe let's throw them in, and then the first time one needs to be added ALONG with updating a snap, we rip that one out? Does that seem like a decent compromise?

I guess, sure. The main concern i had was bloat, as we're checking the same thing twice, although this compromise seems fine.



> (I didn't get rid of any right?)

No i dont think you removed any. 

For tests where we, for example, check stuff from a toml file, can't we parse the toml file and then compare that way? seems less error prone to me

jlizen: I have no objection to switching to that, but I think we accomplish our main goals via simple substr checks. Glad to stamp a PR adding that in! (No time now to add it though, getting on a plane)

nikomatsakis: I think the simple substring checks is fine myself, parsing feels like overkill. 

My experience from multiple projects is

* you want the full layout check, definitely
* but when reviewing diffs it can be quite hard to remember the *point* of the test, so if you can add a few lines like "it really ought to say the word `foo` somewhere" you help prevent bugs -- but if the thing you are testing is just "did it give generally the right shape of output", that's not useful, because it'll be easy to spot. It's more like "if it goes from a list of 5 things to a list of 4 things, is that a problem or is that expected".

nikomatsakis (approved):
r=me

jlizen:
@JayanAXHF mind taking a look whenever you have a moment? currently the previous review is blocking merge. Glad to make other adjustments if you have specific things in mind, I've lost track

JayanAXHF:
Sure, gimme a day or two

JayanAXHF:
I'm fine with the compromise you suggested, i think its a good middle ground. Thanks

JayanAXHF (approved):

Issue #109, opened 2026-04-21
[ci] Apply best practices from CI battery pack


jlizen:
#101 added a CI battery pack with all sorts of best practices we are diverging from. We should modernize our CI a bit!

Specifically:
- split apart our one giant built task that runs test into build, lint, test, etc
- add clippy-sarif for nice inline comments
- msrv validation
- feature powerset
- rust audit canary not running in PRs (and other things in the canary)

PR #112, opened 2026-04-22
feat: merge templates into existing project

### LINKED ISSUE #107: [feat] Merge templates into existing project
We are seeing a growing number of "micro" templates that deal with a specific area of your project - setting up CI (#101 ), adding typed errors, etc. Where a user might want to bootstrap an "entire" project using the CLI templates, these ones are more additive.

Today, the way to use them with existing projects is to use the "preview" feature and copy them over. This is ok, but not ideal. Instead, it would be great to support eg `cargo bp use ci trusted-publishing` that modifies your project in place.

There are some questions to answer (handling collisions, merging into existing `Cargo.toml`, etc). But, they seem solvable.

This runs parallel to a similar discussion happening upstream with module-level cargo examples: https://internals.rust-lang.org/t/alternative-to-cargo-new-templates-examples-as-build-target-templates/24047

I think that `cargo-bp` is a good place to prototype this sort of functionality. It is clearly useful for existing BPs.

jlizen:
Closes https://github.com/battery-pack-rs/battery-pack/issues/107

### Summary

Adds `cargo bp add <pack> -t <template>`, which renders a battery pack template and merges the output into the current project. This makes "micro" templates (spellcheck, fuzzing, mdbook, etc.) usable without manually copying files from a preview.

```bash
cargo bp add ci -t spellcheck
cargo bp add ci -t fuzzing -d ci_platform=github
cargo bp add ci -t trusted-publishing
cargo bp add ci -t benchmarks -N
```

### How it works

Merge strategy depends on file type:

| File type | New file | Existing file |
|-----------|----------|---------------|
| `Cargo.toml` | Write | Merge: deps upgraded if behind, missing features added |
| Other `.toml` files | Write | Merge: new sections and keys added, existing ones left alone |
| `.yml` / `.yaml` files | Write | Merge: top-level keys added additively, existing keys left alone |
| Everything else | Write | Prompt to skip, overwrite, or view diff |

Conflict prompts use single-key shortcuts (`[a]ccept`, `[s]kip`, `[e]dit`, `[A]ccept all`, `[S]kip all`). For TOML/YAML merges, the user can also open the merged result in `$EDITOR` before accepting. Diff output is colored (green/red/cyan).

Files that merge to identical content show as "unchanged" (dimmed), distinct from "skipped" (user chose to skip a conflict).

Flags:
- `--overwrite`: force overwrite non-TOML/YAML files. TOML and YAML always merge, never overwrite.
- `-N`: skip non-TOML/YAML conflicts (unless `--overwrite`). TOML/YAML merges apply automatically.
- `-d key=value`: set template variables without prompting.

Before merging, warns if the git working tree is dirty. In `-N` mode, refuses unless `--overwrite` is passed.

`project_name` is inferred from `Cargo.toml` `[package].name`, falling back to the directory name.

Template authors can add `[[hints]]` to `bp-template.toml` for post-merge instructions:

```toml
[[hints]]
message = "Add `mod errors;` to your lib.rs or main.rs"
```

The TUI detail view gets a `u` keybinding ("Use in project") that triggers the same merge flow. `--path` and `--crate-source` are propagated to the spawned command.

New dependencies: `yaml-rust2`, `similar`.

Also in this PR:
- User docs split into `using.md` (dependency management) and `templates.md` (template operations). Fixes broken cross-references in `intro.md` and `getting-started.md`.
- CI battery pack templates now include `[[hints]]` for post-merge follow-up (tool installation, file editing, workspace config).
- `CONTRIBUTING.md` gets a development section (running dev builds, tests, snapshots).

### Testing

The unit tests you'd expect - toml merging, yaml merging, overwrite, non-interactive, tui screens, snaps of CLI, etc.

Also did a bunch of e2e testing with both CLI and TUI against an existing project. Seems reasonable. I can follow up with a PR to do some of #109 .

### Design decisions

**`add -t` instead of a new `use` subcommand.** `add` already handles battery pack operations on existing projects. `-t` keeps the CLI surface smaller and mirrors the existing `-t` on `new` and `show`. No ambiguity with the positional `<crate>` args.

**Per-file prompts instead of `--dry-run`.** The user sees each conflict in context and decides immediately. `--overwrite` covers the scripting case.

**YAML merge via parser round-trip.** String-level insertion would preserve comments but is fragile for nested structures. The common case (new workflow file, no conflict) doesn't hit the reformatting issue. When it does, a diff is shown.

**Hints instead of auto-inserting `mod` declarations.** Auto-insertion is fragile across module styles (`mod.rs` vs flat, `#[path]`, conditional compilation). `[[hints]]` covers more than modules (tool installation, config steps) and can be upgraded later.

**No `merge` flag in `bp-template.toml`.** Any template works with `add -t`. No need for the author to opt in.

### Known limitations

- **YAML merge reformats files.** Comments are lost and indentation may change. A diff is shown before applying.
- **YAML deep merge is limited to `jobs`, `on`, `permissions`.** Other top-level keys are atomic (existing value wins). This covers GitHub Actions workflows but may need expansion for other YAML formats.

### Intentionally out of scope

- **Relocatable module paths.** Letting template authors declare `.rs` files as movable (`[[modules]]` with a prompt for destination path). Useful for templates that produce source files the user wants to place in their own module tree, but it's a spec change that interacts with hints.
- **Auto-insert `mod` declarations.** Upgrading `[[hints]]` to automatically add `mod foo;` to a target file. Needs `syn` to find the right insertion point.
- **String-level YAML merge.** Detect what to add via the parser, but splice raw text to preserve comments and formatting.
- **`cargo bp validate` in merge mode.** Test templates by merging into a fixture project, not just generating a fresh one.
- **Workspace-aware template merge.** Route template deps through `[workspace.dependencies]`.
- **Nicer diff experience** Right now if you choose "edit", it jumps you into a file view with merge already applied. Ideally we could get block by block or something nicer.
- **Conditional [[hints]]** Currently hints are displayed unconditionally after using a template. Ideally we could branch on whether optional features are selected - for instance the ci battery pack's "all" template, has many optional features that need their own hints conditionally.


md/creating.md
```diff
@@ -314,3 +314,34 @@ mod tests {
...
+- **`Cargo.toml`**: dependencies are merged (versions upgraded if behind, missing features added)
+- **Other `.toml`**: new sections and keys added, existing ones left alone
+- **YAML**: top-level keys merged additively (new jobs added, existing ones preserved)
+- **Other files**: the user is prompted to skip or overwrite if they already exist
+
+Note: YAML merges don't preserve comments in existing files. For YAML, only `jobs`, `on`, and `permissions` are deep-merged; other top-level keys are atomic, so if the key already exists, the user's value wins. To avoid conflicts, use unique workflow filenames (e.g., `typos.yml` instead of `ci.yml`).
+
+To help users with steps that can't be automated (like adding `mod` declarations or installing tools), declare hints in your `bp-template.toml`:
+
+```toml
```
nikomatsakis, line 333: I wonder if we could add an optional mode where we use ACP to dispatch this to an agent to do.

nikomatsakis (approved):
This looks quite good. I like the idea of a fancier TUI that lets you preview the changes and move back and forth, but for a first pass, this is awesome. I gotta try it out!

md/creating.md
jlizen: Yeah we should keep iterating on this "follow up tasks" workflow

PR #106, opened 2026-04-19
Remove build.rs hooks, add cargo bp check for drift detection

nikomatsakis:
- Remove validate functions from battery packs (were just TODOs)
- Add `cargo bp check` command with version drift detection
- Check for outdated versions vs battery pack recommendations
- Clean implementation without build.rs pollution

PR #117, opened 2026-04-28
Remove build.rs hooks, add cargo bp check

nikomatsakis:
This PR removes the automatic build.rs creation that was forcing all battery pack users to have build scripts in their projects. Previously, running `cargo bp add cli` would create or modify a build.rs file with validation calls, which was intrusive and unnecessary for most users.

**Key Changes:**
- Remove automatic build.rs creation from `cargo bp add`
- Remove build-dependencies registration for battery packs  
- Add new `cargo bp check` command for optional drift detection
- Remove validate functions from battery packs (were just TODOs)

**Benefits:**
- Clean user experience without forced build scripts
- Optional validation via `cargo bp check` when needed
- Simpler codebase with 100+ fewer lines of code
- Battery packs tracked via metadata instead of build-dependencies

Users who want validation can run `cargo bp check` on demand to see version drift warnings, while everyone else gets a cleaner experience without build.rs pollution.

### DIFF
--- a/src/battery-pack/bphelper-cli/src/commands/mod.rs
+++ b/src/battery-pack/bphelper-cli/src/commands/mod.rs
@@ -1403,6 +1403,10 @@ fn status_battery_packs(
     Ok(())
 }
 
+/// Check installed battery packs for version drift.
+/// 
+/// Compares user's current dependency versions against battery pack recommendations
+/// and warns when user versions are older than recommended versions.
 fn check_battery_packs(
     project_dir: &Path,
     _path: Option<&str>,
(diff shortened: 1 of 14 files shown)


src/battery-pack/bphelper-cli/src/commands/mod.rs
```diff
@@ -2021,50 +1406,41 @@ fn status_battery_packs(
 fn check_battery_packs(
```
nikomatsakis, line 1410: please add a comment explaining the purpose of this function

nikomatsakis:
Closing as duplicate - these exact changes were already merged in PR #106 on April 19, 2026. The build.rs removal and cargo bp check command are already implemented in main.

Issue #13, opened 2026-03-01
Consider using `cargo_metadata` rather than hand rolled parsing


jlizen:
Currently we directly parse `Cargo.toml` for workspace member discovery, parsing crate dependencies, etc.

https://github.com/battery-pack-rs/battery-pack/blob/a69a1e7f2990bf3b6b18562561e9b83c0ca0a561/src/battery-pack/bphelper-manifest/src/lib.rs#L797 

This is nice in that:
- we don't get extra noise like implicit features per dependency that we don't want to expose to bp consumers
- we don't need a separate pass to handle our custom metadata tables/fields
- it keeps tests simple since we don't need a subprocess call against a materialized Cargo workspace (as required by `cargo_metadata`)

But, it almost certainly have gaps on the workspace discovery side compared to full cargo spec. And then, on the parsing side, we probably should align with cargo's approach, even if there is some extra code to filter down what bp consumers see, if we want to eventually unify this feature with cargo handling somehow.

So, it's probably a good idea to cut to [cargo_metadata](https://crates.io/crates/cargo_metadata).

Per @epage :
> Almost all cargo tests are end-to-end tests, using `cargo-test-support` to have a per-test registry, local packages, git packages, etc
If this is ever upstreamed, that will be the expectation.

gmemuriuki:
@jlizen I'll work on this - can you assign it to me?

PR #118, opened 2026-04-28
refact: store the data in battery-pack.toml

nikomatsakis:


src/battery-pack/bphelper-cli/src/commands/mod.rs
```diff
@@ -109,8 +109,7 @@ pub(crate) enum BpCommands {
         all_features: bool,
 
         // [impl cli.add.target]
-        /// Where to store the battery pack registration
-        /// (workspace, package, or default)
+        /// Compatibility flag; state is stored in `battery-pack.toml`.
```
nikomatsakis: We don't need compatibility flags here. I think we can just remove this.

src/battery-pack/bphelper-cli/src/commands/mod.rs
```diff
@@ -215,11 +214,11 @@ pub(crate) enum BpCommands {
...
 #[derive(Debug, Clone, Copy, PartialEq, Eq, clap::ValueEnum)]
 pub(crate) enum AddTarget {
-    /// Register in `workspace.metadata.battery-pack`.
+    /// Deprecated compatibility value.
     Workspace,
-    /// Register in `package.metadata.battery-pack`.
+    /// Deprecated compatibility value.
     Package,
-    /// Use workspace if a workspace root exists, otherwise package
+    /// Deprecated compatibility value.
```
nikomatsakis: and these

src/battery-pack/bphelper-cli/src/commands/mod.rs
```diff
@@ -379,6 +381,20 @@ pub fn main() -> Result<()> {
...
 }
 
+/// Preflight: keep `battery-pack.toml` managed-deps aligned with current Cargo.toml.
+pub(crate) fn sync_state_with_current_manifest(project_dir: &Path) -> Result<usize> {
+    let user_manifest_path = match find_user_manifest(project_dir) {
+        Ok(path) => path,
+        Err(_) => return Ok(0),
+    };
+    let user_manifest_content =
+        std::fs::read_to_string(&user_manifest_path).context("Failed to read Cargo.toml")?;
```
nikomatsakis, line 398: Hmm. Don't we use `cargo metadata` to read the manifest?

src/battery-pack/bphelper-cli/src/commands/mod.rs
```diff
@@ -697,7 +713,7 @@ pub(crate) fn add_battery_pack(
     no_default_features: bool,
     all_features: bool,
     specific_crates: &[String],
-    target: Option<AddTarget>,
+    _target: Option<AddTarget>,
```
nikomatsakis: remove

src/battery-pack/bphelper-cli/src/commands/mod.rs
```diff
@@ -1286,6 +1218,112 @@ fn sync_battery_packs(project_dir: &Path, path: Option<&str>, source: &CrateSour
     Ok(())
 }
 
+fn enable_feature(
```
nikomatsakis: This...appears to be dead code?

src/battery-pack/bphelper-cli/src/commands/mod.rs
JayanAXHF: Using `cargo metadata` was suggested in #13 but hasn't been implemented yet

JayanAXHF (commented):
So, is the `battery-pack.toml`created on the workspace level, crate level, or smth else?

src/battery-pack/bphelper-cli/src/commands/mod.rs
JayanAXHF: we can always remove it and put a comment there just in case

src/battery-pack/bphelper-cli/src/commands/mod.rs
```diff
@@ -379,6 +363,34 @@ pub fn main() -> Result<()> {
...
+        .current_dir(project_dir)
+        .no_deps()
+        .exec()
+        .context("Failed to run `cargo metadata`")?;
+
+    // With --no-deps the resolve graph is absent; find the root package
+    // via workspace_members instead.
+    let package = metadata
+        .workspace_members
+        .first()
```
jlizen: This will not behave correctly in a multi-package workspace. `current_dir` affects where cargo's search for the workspace starts, but it still walks upwards until it reaches workspace root. 

This means that  `cargo metadata` with `--no-deps` returns all members of the virtual workspace regardless of what directory it is passed. So this would return the first workspace member based on (currently alphabetical, but nonspecified) cargo ordering, instead of whatever is in the current directory.

Instead, we would want something like:
```
  let project_dir = project_dir
      .canonicalize()
      .unwrap_or_else(|_| project_dir.to_path_buf());

  let package = metadata
      .packages
      .iter()
      .find(|p| {
          p.manifest_path
              .parent()
              .and_then(|d| d.canonicalize().ok())
              == Some(project_dir.clone())
      });
```

src/battery-pack/bphelper-cli/src/commands/mod.rs
jlizen: (Would be great to do though if you felt like some ol' fashioned scope creep)

src/battery-pack/bphelper-cli/src/commands/mod.rs
```diff
@@ -1065,20 +1036,11 @@ fn remove_battery_pack(
...
-            .and_then(|m| m.get_mut("battery-pack"))
-            .and_then(|bp| bp.as_table_mut())
-    {
-        bp_table.remove(&crate_name);
-    }
-
     std::fs::write(&user_manifest_path, user_doc.to_string())
         .context("Failed to write Cargo.toml")?;
 
+    let _ = remove_battery_pack_state_entry(&user_manifest_path, &crate_name)?;
```
jlizen: Probably should `eprintln!()` a warning since this will leave behind a stale `battery-pack.toml`?

src/battery-pack/bphelper-cli/src/manifest/mod.rs
```diff
@@ -506,73 +628,140 @@ pub(crate) fn read_managed_deps_from(
...
+
+/// Upsert battery-pack state in sibling `battery-pack.toml`.
+pub(crate) fn write_battery_pack_state(
+    user_manifest_path: &Path,
+    bp_name: &str,
+    active_features: &BTreeSet<String>,
+    managed_crates: &BTreeMap<String, bphelper_manifest::CrateSpec>,
+) -> Result<()> {
+    let state_path = state_file_path(user_manifest_path);
+    let mut state = read_state_file(&state_path)?;
```
jlizen, line 542: Minor: We have this same pattern (read state / modify / serialize / add newline / flush) in three places (`write_battery_pack_state`, `remove_battery_pack_state_entry`, `prune_state_managed_deps_for_manifest`). Worth a helper?

src/battery-pack/bphelper-cli/src/manifest/mod.rs
```diff
@@ -355,6 +356,114 @@ pub(crate) fn sync_dep_in_table(
...
+    managed_deps: Vec<ManagedDepEntry>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+struct ManagedDepEntry {
+    name: String,
+    version: String,
+}
+
+fn short_pack_name(bp_name: &str) -> String {
```
jlizen: Can we dedupe this with `crate::registry::short_name` (or just use that directly)?

src/battery-pack/bphelper-cli/src/commands/mod.rs
jlizen: The comment also refers to a "root" package, which I don't think is quite what we want... for one, that doesn't exist in a virtual workspace.

And then anyway even in a non-virtual workspace, it looks like want the `battery-pack.toml` to live next to each package's `Cargo.toml`, rather than on the root package.

If I'm understanding that intent correctly, it would only work properly for a single-package workspace.

src/battery-pack/bphelper-cli/src/manifest/mod.rs
```diff
@@ -382,10 +491,13 @@ pub(crate) fn read_features_at(
 }
 
 /// Read active features for a battery pack from user's package metadata.
+///
+/// This is a legacy reader kept for backward compatibility with projects
+/// that still store state in `Cargo.toml` metadata.
```
jlizen: It also might be worth thinking about doing some automatic cleanup for "legacy" metadata to move it into the new setup, if we are going to the trouble of supporting the old usage. Right now it just will keep using the old metadata until we someday rip out support.

Probably also worth adding some tests validating that this backwards compat still works.

jlizen (changes requested):
Main concern is how we are handling resolving manifests in multi-package workspaces, see [this comment](https://github.com/battery-pack-rs/battery-pack/pull/118/changes#r3163760993). We should also have some tests for virtual and non-virtual multi-package workspaces.

It would also be good to update the mdbook to describe the new format, once this comes out of draft. It's a bit unclear about our intent here (does `battery-pack.toml` live next to each crate or at workspace root? I think next to?)

src/battery-pack/bphelper-cli/src/manifest/mod.rs
jlizen: That said, I think we could probably safely ignore the old code approach entirely, I don't think we have non-toy consumers yet

src/battery-pack/bphelper-cli/src/commands/mod.rs
```diff
@@ -379,6 +363,38 @@ pub fn main() -> Result<()> {
...
+    let project_dir = project_dir
+        .canonicalize()
+        .unwrap_or_else(|_| project_dir.to_path_buf());
+
+    let package = metadata.packages.iter().find(|p| {
+        p.manifest_path.parent().and_then(|d| d.canonicalize().ok()) == Some(project_dir.clone())
+    });
+
+    let Some(package) = package else {
+        return Ok(0);
```
jlizen, line 393: would be good to have a trace-level log here for debug

src/battery-pack/bphelper-cli/src/manifest/mod.rs
```diff
@@ -355,6 +356,134 @@ pub(crate) fn sync_dep_in_table(
...
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+struct ManagedDepEntry {
+    name: String,
+    version: String,
+}
+
+use crate::registry::short_name;
+
+fn state_name_matches(name: &str, bp_name: &str) -> bool {
```
jlizen, line 413: this will fall over if passed `cli-battery-pack` (which currently the CLI shouldn't, but it seems unnecessarily confusing).

might as well normalize with `short_name()` first

src/battery-pack/bphelper-cli/src/manifest/mod.rs
```diff
@@ -355,6 +356,134 @@ pub(crate) fn sync_dep_in_table(
...
+
+const STATE_FORMAT_VERSION: u32 = 1;
+
+fn default_version() -> u32 {
+    STATE_FORMAT_VERSION
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+struct BatteryPackStateFile {
+    #[serde(default = "default_version")]
```
jlizen, line 381: Might be cleaner to just impl Default for BatteryPackStateFile so we don't derive it with 0?

jlizen (approved):
some nits, i'll push a quick patch up per our offline discussion

PR #122, opened 2026-04-30
refactor: Rename validate_templates to validate, improve bp-managed, fix version drift

jlizen:
### Summary

The recent tarball validation change (#121) exposed a version drift problem: the default authoring template hardcodes `battery-pack = "0.4"`, which falls behind as new versions are released. The fix is to use `battery-pack.bp-managed = true` instead, but that required two prerequisite changes to bp-managed resolution. This PR chains those together.

I also fixed an issue when the new validation runs in a release-plz step due to how Cargo packaging works (it expects all locally referenced versions to be in the index). Normally release_plz only calls `cargo package` in the proper order AFTER publishing earlier dependencies, where our tests run BEFORE that step.

### CI fix from the new validation

During multi-crate releases (e.g. via release-plz), workspace path dependencies may be bumped to versions not yet published to crates.io, causing `cargo package` to fail. Validation now detects this case and falls back to validating directly from the source tree. This is less precise (it won't catch files excluded from the tarball) but still catches template render and build errors.

### Rename `validate_templates` to `validate`

The function now validates more than just templates (it packages the crate and validates from the tarball), so the shorter name better reflects its scope. All consumer test functions and documentation are updated. My hope is that the shorter name will nudge authors to keep the validation around even without templates, which saves them from significant future confusion if they later add one and mix up the `_Cargo.toml` naming.

Making this change surfaced drift in our authoring templates' version (since the validation API name started to mismatch), which the next two sections address.

### Improve `bp-managed` resolution.

The authoring template needs `battery-pack = { bp-managed = true, features = ["cli"] }` in its dev-dependencies, but previously `bp-managed = true` rejected any other keys on the dependency entry. So, it hardcoded the version instead, which caused the drift in the next item. Now only `version` conflicts (since `bp-managed` provides the version). Keys like `features`, `optional`, and `default-features` are preserved in the resolved output. If no explicit features are given, the spec's features are used as defaults. If explicit features are present, they override the spec entirely.

Additionally, battery packs can now reference themselves via `bp-managed`. The `discover_battery_packs` filter was updated to include `battery-pack` itself (not just `*-battery-pack`), and the resolution fallback registers all discovered battery pack versions so self-references resolve correctly. That was also forcing hardcoding.

### Fix version drift in the default template.

With the above in place, the authoring template's `_Cargo.toml` now uses `battery-pack.bp-managed = true` instead of the hardcoded version. The `with_template` authoring template was rewritten to use its own `_Cargo.toml` (instead of `{% include %}` from the default template).

### Add snapshot tests to the authoring template.

In the process of all of these fies, I found it pretty useful to have good snapbox testing for the ci templates. Since this PR anyway touches the authoring templates, throwing in snapbox usage in the `with_template` template.

### Testing

The bp-managed resolution changes are covered by new unit tests for explicit feature override, `default-features = false` passthrough, battery pack self-reference, and the `version` conflict error. The `discover_battery_packs` filter fix is verified by a test that discovers `battery-pack` from its own crate root.

The with_template integration test validates two-level generation end-to-end: generating a battery pack from the authoring template, then validating that the generated battery pack's inner template produces a working project. Snapshot tests verify both the outer Cargo.toml (with resolved bp-managed versions) and the inner lib.rs content. And then, the `validate()` call runs the templates' own tests, which themselves check their own snapshots.

The CI fix (source tree fallback) is covered by unit tests that verify the stderr pattern matching: workspace dependency failures are detected and skipped, while external dependency failures are still treated as errors. An integration test verifies that a template file excluded via `package.exclude` is caught by the tarball validation path. The fallback was also verified manually by bumping workspace dependency versions to reproduce the failure mode seen in [this CI run](https://github.com/battery-pack-rs/battery-pack/actions/runs/25180848032/job/73825374849).


src/battery-pack/bphelper-cli/src/validate/tests.rs
```diff
@@ -210,3 +210,78 @@ test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out;
 "#]]
     );
 }
+
+#[test]
```
jlizen, line 214: This "tests.rs next to the main module" pattern is bad. #123 to fix it in a time that won't cause merge conflicts.

src/battery-pack/templates/with_template/templates/default/build.rs
jlizen, line 1: This survived our tear-out of build.rs tendrils during `cargo bp new` (though this one wasn't set up right anyway)

nikomatsakis (approved):

PR #128, opened 2026-05-11
feat: ci-battery-pack skills

jlizen:
Adding some agent spec skills to our `ci-battery-pack`, designed to integrate with `symposium`.

Currently we have lots of good templates, but the related crates.io/github secrets config can be a bit fiddly. Having skills that spell things out more is nice.

Here you can see a (n=1) eval using these skills: [eval link](https://gist.github.com/jlizen/45f65ceca392ae387448db6b8d5cabbb)

Unfortunately we don't have a great way to tell symposium to load this until https://github.com/symposium-dev/symposium/issues/210 lands, but getting skills in is a good start. We could also do something on our own end with some other `cargo bp skills`-like command, but I don't want to tackle that here. My suspicion is that we'd be happier just delegating skill loading to symposium.


jlizen:
I intentionally did NOT list out all the possible placeholders that our templates can switch on. Probably we should for maximum usability, but I think it's a maintenance nightmare.

Instead, I'd rather land https://github.com/battery-pack-rs/battery-pack/issues/131 and then add a brief mention of calling `cargo bp show ci -t full --help` to see all available flags.

nikomatsakis (changes requested):
let's rename this to `.symposium/skills`

jlizen:
- Moved `battery-packs/*/skills` -> `battery-packs/*/.symposium/skills`
- Removed all `SYMPOSIUM.toml` since we rely on the `recommendations` repo and are setting up a different system to allow dogfooding. I can tweak the benchmarks later to use whatever new system, but whoever uses this next should be able to hack it together if needed.
- Remove mention of `CARGO_BP_NON_INTERACTIVE=true`, it was just noise since anyway we detect non-TTY.

nikomatsakis:
I think we should merge these butI also think I want to change back from `.symposium/skills` to `skills`.

jlizen:
Updated the naming @nikomatsakis (also friendly plug for #136 ^^)

I can't merge on my own due to the current "changes requested" flag from your last review, fyi.

nikomatsakis (approved):

Issue #103, opened 2026-04-19
[feat] Improve template engine performance for pin_github_action


jlizen:
Added in #101 , `pin_github_action()` resolves GitHub Action SHAs via `git ls-remote` at render time. This works but is slow on first use (~2-3s) and blocks the TUI.

### Proposed improvements

1. **Eager background prefetch** — start resolving SHAs when the battery pack list screen loads, not when the user opens a preview. By the time they navigate to preview, results are cached.

2. **Async template engine** — the engine makes multiple network calls (pin_github_action, crates.io for bp-managed). These should be async rather than blocking threads.

3. **Disk cache with TTL** — cache resolved SHAs to `<NOT SURE EXACTLY WHERE>` with ~1 day TTL. Repeated runs within a day skip network entirely. It makes me a bit jumpy to have a TTL since what if there is a supply chain attacked version we don't want to resolve to... Probably should think about this a bit more before doing anything.

### Current state

- Parallel prefetch within a single render (threads, not async)
- Global in-memory session cache (persists across preview navigations)
- No disk persistence

PR #130, opened 2026-05-11
fix(tui): make Enter on template show preview instead of opening browser

jlizen:
Ref: https://rust-lang.zulipchat.com/#narrow/channel/575623-battery-packs-rs/topic/publicity/near/594416142

Two fixes:

## Load crate source for preview from within tui

Currently, this properly jumps into the tui with template materialized;
```
cargo bp show ci -t full
```

However this:
```
cargo bp show ci
```
followed by hitting "p" (preview) on the template shows:

```
 Template preview unavailable — battery pack not found locally.
 Use --crate-source or install the pack first.
```

This is a strictly worse experience, updating to also actively download for the tui navigation if needed.

## "return" on templates opens preview

Currently, "return" attempts to open them in github tree (or if on a headless server, generally just fails). Opening a browser makes more sense for inspecting individual crates, but much less for templates.

Since, templates have (potentially complex, branching) jinja syntax and are generally not optimized for human readability. Instead, we can just serve the preview, which is already materialized and works in more environments.

The new behavior is, "return" opens the preview for templates (but still opens the browser for everything else). And, i left in place a redundant "p" for "preview" for symmetric with `cargo add`, where hitting "return" does something different (add template to your project).

A side benefit of this is that it let us remove a bunch of code :) 

## Testing

Went through the worfklows myself, manually.

We don't have great automated testing of our tui navigation and I don't have it in me to add now. The view was already well-tested, so there is no snap to change, really.

### DIFF
--- a/src/battery-pack/bphelper-cli/src/registry/mod.rs
+++ b/src/battery-pack/bphelper-cli/src/registry/mod.rs
@@ -146,9 +146,6 @@ pub(crate) struct TemplateInfo {
     pub name: String,
     pub path: String,
     pub description: Option<String>,
-    /// Full path in the repository (e.g., "src/cli-battery-pack/templates/simple")
-    /// Resolved by searching the GitHub tree API
-    pub repo_path: Option<String>,
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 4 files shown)


jlizen:
@nikomatsakis I addressed most of the feedback from copilot (pretty good), including:
- adding a state transition test
- fixing some stale doc comments
- lazily getting repo path only if there are examples
- error message improvements

I did NOT take the suggestion around adding a "loading" screen or other things to have a nicer experience than synchronously loading the referenced template. We do have a minor UI delay as a result. That is beyond the scope of this PR. Some options are discussed to some extent in #103 , I think a similar set of solutions could apply.

nikomatsakis:
Merging.

PR #134, opened 2026-05-19
Fix Windows path handling in template validation 

gmemuriuki:
## Problem
The template scaffolding logic had two Windows path-handling bugs that broke battery-pack::test::validate on windows

- The _Cargo.toml -> Cargo.toml rename used string-based starts_with("templates/") to skip nested scaffolds. Windows path use \ separators, so the check always failed, the rename ran inside templates/, and cargo treated the result as a subpackage, which packaging then dropped.

- The workpace path injection wrote TOML basic strings ("{path}"). Windows paths like C:\users\... contains \U, which TOML parsers rejects as invalid escapes.

### Fixes

1. template_engine.rs -- switch _Cargo.toml rename guard to Path::starts_with(separator-agnostic).
2. write_crates_io_patches -- emit TOML literal strings ('{path}' no escape processing) instead of basic strings.

Add .claude/ to .gitignore.

**battery-pack::test::validate on Windows for the with_template fixture.**


.gitignore
```diff
@@ -2,3 +2,4 @@ book
 target
 WIP.md
 .claude/settings.local.json 
+.claude/
```
jlizen: it looks like we do have at least one intentionally committed prompt in `.claude/`, maybe we could just add `.claude/plans` to `.gitignore` instead? (and wipe out that lingering plan that found its way in)

I assume this change was prompted by accidental inclusions via symposium, but newer versions should be self-gitignoring, right?

jlizen (commented):
Overall change looks good. Could we add Windows to the test matrix though so that this sort of regression is properly captured in CI? Probably darwin too while we're at it.

Eg:
```
    test:
      strategy:
        matrix:
          os: [ubuntu-latest, windows-latest, macos-latest]
      runs-on: ${{ matrix.os }}
      steps:
        - uses: actions/checkout@v4
        - uses: Swatinem/rust-cache@v2
        - name: Build
          run: cargo build
        - name: Run tests
          run: cargo test --verbose
```

gmemuriuki:
Sure 💯

.gitignore
gmemuriuki: 😄yes.

> I assume this change was prompted by accidental inclusions via symposium, but newer versions should be self-gitignoring, right?

Yes

gmemuriuki:
Now I have more tests failing (not because of my changes) but existing window-specific issues in the code. I'll fix them first, so please don't merge this right now.

JayanAXHF:
Please ping when this is ready to merge, thanks

jlizen (approved):
Looks good, pushing up some minor tweaks then will merge.

jlizen:
@gme-muriuki this is ready to merge, i think you can push directly using the merge when ready button but feel free to ping me if not

(sorry to push up directly while you were still working on it, didn't realize, was trying to save you a bit of churn ^^)

gmemuriuki:
> @gme-muriuki this is ready to merge, i think you can push directly using the merge when ready button but feel free to ping me if not

Yeah, it is. I have no more changes to push here.
 
> (sorry to push up directly while you were still working on it, didn't realize, was trying to save you a bit of churn ^^)

That's so sweet. Thank you.

Issue #96, opened 2026-04-18
[bug] `cargo bp add --path` should resolve `workspace = true` dependencies


jlizen:
Related: #13 

`parse_single_dep` in `bphelper-manifest` reads the version directly from the raw `Cargo.toml`. Deps using workspace inheritance have no `version` key, so they resolve to `""`:

```toml
# battery pack Cargo.toml
[dev-dependencies]
snapbox = { workspace = true, features = ["cmd"] }
```

```toml
# written to user project by cargo bp add --path
[dev-dependencies]
snapbox = ""
```

Resolving these via `cargo metadata` would let battery packs use workspace deps without breaking the local `--path` workflow.

gmemuriuki:
@jlizen I would love to work on this issue.

Also, does related to #13 mean that the issue (13) needs to be resolved before I start working on this one?

jlizen:
My advice would be to do #13 first and it might fix this for free, but there is not something fundamentally wrong with starting here, it just might be more of a patchwork fix.

gmemuriuki:
I'll tackle #13 first since it might fix this issue too. If not, I'll address this one separately.

gmemuriuki:
I think this PR closed this issue here https://github.com/battery-pack-rs/battery-pack/pull/135. @jlizen

PR #135, opened 2026-05-19
refactor(manifest): use cargo_metadata for parsing and discovery

gmemuriuki:
## Problem
`bhelper-manifest` used a hand-rolled TOML parser that walked workspace members via literal path joins.
This missed glob members, default-members, exclude, and workspace dependency inheritance - mostly visibly, `snapbox = {workspace = true}` resolved to an empty version, so cargo bp add --path wrote `snapbox = ""` (#96).

## Fix
- Replace the parser with cargo_metadata::MetaCommand, which shells out to cargo metadata and returns Cargo's own dependency-resolved view.

## Changes:
1. Delete parse_battery_pack(&str), parse_dep_section, parse_single_dep, and the Raw* deserialization types.
2. Add `parse_battery_pack_from_path(&Path)` as the single-pack entry point; migrate 16 call sites.
3. Collapse discover_from_crate_root into `discover_battery_packs` -- cargo metadata --manifest-path walks to the workspace root, and a no-[workspace] crate is a 1-member workspace, so both paths are identical.
4. Extract load_metadata for shared MetadataCommand boilerplate.
5. Side-read `[package.metadata.battery-pack]` and `[package.metadata.battery.templates]` from TOML directly -- cargo_metadata exposes these as serde_json::Value which loses the original shape.
6. Switch completion cache from raw Cargo.toml snippets to JSON-serialized BatteryPackSpec -- cargo metadata can't operate on isolated manifests.

**Closes (#13) and (#96)**

### DIFF
--- a/src/battery-pack/bphelper-cli/src/commands/mod.rs
+++ b/src/battery-pack/bphelper-cli/src/commands/mod.rs
@@ -3,7 +3,7 @@
 //! This module contains the `main()` entry point and all subcommand handlers.
 //! Depends on `registry` and `manifest`.
 
-use anyhow::{Context, Result, anyhow, bail};
+use anyhow::{Context, Result, bail};
 use bphelper_manifest::parse_battery_pack_from_path;
 use clap::{CommandFactory, Parser, Subcommand};
 use std::collections::{BTreeMap, BTreeSet};
@@ -746,10 +746,9 @@ pub(crate) fn add_battery_pack(
     let (bp_version, bp_spec) = if let Some(local_path) = path {
         let manifest_path = Path::new(local_path).join("Cargo.toml");
         let spec = parse_battery_pack_from_path(&manifest_path)
-            .map_err(|err| anyhow!("Failed to parse battery pack '{}': {}", crate_name, err))?;
-        let version = spec.version.clone(); // !quite sure?
+            .with_context(|| format!("Failed to parse battery pack '{}'", crate_name))?;
 
-        (Some(version), spec)
+        (None, spec)
     } else {
         fetch_bp_spec(source, name)?
     };
@@ -1556,8 +1555,7 @@ fn parse_template_metadata(
     manifest_path: &Path,
     crate_name: &str,
 ) -> Result<BTreeMap<String, TemplateConfig>> {
-    let spec = parse_battery_pack_from_path(manifest_path)
-        .map_err(|err| anyhow!("Failed to parse Cargo.toml: {}", err))?;
+    let spec = parse_battery_pack_from_path(manifest_path).context("Failed to parse Cargo.toml")?;
 
     if spec.templates.is_empty() {
         bail!(
--- a/src/battery-pack/bphelper-cli/src/commands/mod.rs
+++ b/src/battery-pack/bphelper-cli/src/commands/mod.rs
@@ -1309,7 +1309,7 @@ fn pick_crates_interactive(
     use dialoguer::MultiSelect;
 
     // Collect non-default features with their member crates
-    let features: Vec<(&String, &BTreeSet<String>)> = bp_spec
+    let features: Vec<(&String, &BTreeSet<bphelper_manifest::FeatureRef>)> = bp_spec
         .features
         .iter()
         .filter(|(name, _)| name.as_str() != "default")
@@ -1345,8 +1345,8 @@ fn pick_crates_interactive(
     for (feat_name, feat_crates) in &features {
         let member_list = feat_crates
             .iter()
+            .map(|fref| fref.dep_name())
             .filter(|c| !bp_spec.is_hidden(c))
-            .cloned()
             .collect::<Vec<_>>()
             .join(", ");
         labels.push(format!(
@@ -1358,12 +1358,14 @@ fn pick_crates_interactive(
             // Feature is checked if all its visible members are in defaults
             feat_crates
                 .iter()
+                .map(|fref| fref.dep_name())
                 .filter(|c| !bp_spec.is_hidden(c))
                 .all(|c| default_crates.contains(c))
         } else {
             // Feature is checked if all its visible members are pre-selected
             feat_crates
                 .iter()
+                .map(|fref| fref.dep_name())
                 .filter(|c| !bp_spec.is_hidden(c))
                 .all(|c| pre_selected.contains(c))
         };
@@ -1426,9 +1428,10 @@ fn pick_crates_interactive(
         match item {
             PickerItem::Feature(feat_name) => {
                 if let Some(members) = bp_spec.features.get(feat_name) {
-                    for c in members {
+                    for fref in members {
+                        let c = fref.dep_name();
                         if !bp_spec.is_hidden(c) {
-                            selected_crates.insert(c.clone());
+                            selected_crates.insert(c.to_string());
                         }
                     }
                 }
@@ -1454,6 +1457,7 @@ fn pick_crates_interactive(
     if default_members.is_some_and(|members| {
         members
             .iter()
+            .map(|r| r.dep_name())
             .filter(|c| !bp_spec.is_hidden(c))
             .all(|c| selected_crates.contains(c))
     }) {
@@ -1465,6 +1469,7 @@ fn pick_crates_interactive(
         }
         if feat_crates
             .iter()
+            .map(|r| r.dep_name())
             .filter(|c| !bp_spec.is_hidden(c))
             .all(|c| selected_crates.contains(c))
         {
--- a/src/battery-pack/bphelper-cli/src/commands/mod.rs
+++ b/src/battery-pack/bphelper-cli/src/commands/mod.rs
@@ -3,7 +3,8 @@
 //! This module contains the `main()` entry point and all subcommand handlers.
 //! Depends on `registry` and `manifest`.
 
-use anyhow::{Context, Result, bail};
+use anyhow::{Context, Result, anyhow, bail};
+use bphelper_manifest::parse_battery_pack_from_path;
 use clap::{CommandFactory, Parser, Subcommand};
 use std::collections::{BTreeMap, BTreeSet};
 use std::io::IsTerminal;
@@ -456,9 +457,7 @@ fn new_from_battery_pack(opts: NewFromBpOpts<'_>) -> Result<()> {
 
     // Read template metadata from the Cargo.toml
     let manifest_path = resolved.dir.join("Cargo.toml");
-    let manifest_content = std::fs::read_to_string(&manifest_path)
-        .with_context(|| format!("Failed to read {}", manifest_path.display()))?;
-    let templates = parse_template_metadata(&manifest_content, &crate_name)?;
+    let templates = parse_template_metadata(&manifest_path, &crate_name)?;
 
     // Resolve which template to use
     let template_path = resolve_template(&templates, opts.template.as_deref(), opts.interactive)?;
@@ -653,9 +652,7 @@ fn add_template(opts: AddTemplateOpts<'_>) -> Result<()> {
 
     // Read template metadata and resolve which template to use.
     let manifest_path = crate_dir.join("Cargo.toml");
-    let manifest_content = std::fs::read_to_string(&manifest_path)
-        .with_context(|| format!("Failed to read {}", manifest_path.display()))?;
-    let templates = parse_template_metadata(&manifest_content, &crate_name)?;
+    let templates = parse_template_metadata(&manifest_path, &crate_name)?;
     let template_path = resolve_template(&templates, Some(opts.template), opts.interactive)?;
 
     // Load post-merge hints before moving template_path.
@@ -748,11 +745,11 @@ pub(crate) fn add_battery_pack(
     // [impl cli.source.replace]
     let (bp_version, bp_spec) = if let Some(local_path) = path {
         let manifest_path = Path::new(local_path).join("Cargo.toml");
-        let manifest_content = std::fs::read_to_string(&manifest_path)
-            .with_context(|| format!("Failed to read {}", manifest_path.display()))?;
-        let spec = bphelper_manifest::parse_battery_pack(&manifest_content)
-            .map_err(|e| anyhow::anyhow!("Failed to parse battery pack '{}': {}", crate_name, e))?;
-        (None, spec)
+        let spec = parse_battery_pack_from_path(&manifest_path)
+            .map_err(|err| anyhow!("Failed to parse battery pack '{}': {}", crate_name, err))?;
+        let version = spec.version.clone(); // !quite sure?
+
+        (Some(version), spec)
     } else {
         fetch_bp_spec(source, name)?
     };
@@ -1487,14 +1484,12 @@ fn generate_from_local(opts: NewOpts, local_path: &str, template: Option<String>
 
     // Read local Cargo.toml
     let manifest_path = local_path.join("Cargo.toml");
-    let manifest_content = std::fs::read_to_string(&manifest_path)
-        .with_context(|| format!("Failed to read {}", manifest_path.display()))?;
 
     let crate_name = local_path
         .file_name()
         .and_then(|s| s.to_str())
         .unwrap_or("unknown");
-    let templates = parse_template_metadata(&manifest_content, crate_name)?;
+    let templates = parse_template_metadata(&manifest_path, crate_name)?;
     let template_path = resolve_template(&templates, template.as_deref(), opts.interactive)?;
 
     generate_from_path(opts, local_path, &template_path)
@@ -1558,11 +1553,11 @@ fn parse_define(s: &str) -> Result<(String, String), String> {
 }
 
 fn parse_template_metadata(
-    manifest_content: &str,
+    manifest_path: &Path,
     crate_name: &str,
 ) -> Result<BTreeMap<String, TemplateConfig>> {
-    let spec = bphelper_manifest::parse_battery_pack(manifest_content)
-        .map_err(|e| anyhow::anyhow!("Failed to parse Cargo.toml: {}", e))?;
+    let spec = parse_battery_pack_from_path(manifest_path)
+        .map_err(|err| anyhow!("Failed to parse Cargo.toml: {}", err))?;
 
     if spec.templates.is_empty() {
         bail!(
--- a/src/battery-pack/bphelper-cli/src/commands/mod.rs
+++ b/src/battery-pack/bphelper-cli/src/commands/mod.rs
@@ -155,6 +155,10 @@ pub(crate) enum BpCommands {
     List {
         /// Filter by name (omit to list all battery packs)
         filter: Option<String>,
+
+        /// Emit machine-readable JSON instead of the default text output
+        #[arg(long)]
+        json: bool,
     },
 
     /// Show detailed information about a battery pack
@@ -176,6 +180,10 @@ pub(crate) enum BpCommands {
         /// Set a template placeholder value (e.g., -d description="My project")
         #[arg(long = "define", short = 'd', value_parser = parse_define)]
         define: Vec<(String, String)>,
+
+        /// Emit machine-readable JSON instead of the default text output
+        #[arg(long, conflicts_with_all = ["template", "define"])]
+        json: bool,
     },
 
     /// Show status of installed battery packs and version warnings
@@ -301,15 +309,14 @@ pub fn main() -> Result<()> {
                     interactive,
                     &project_dir,
                 ),
-                BpCommands::List { filter } => {
+                BpCommands::List { filter, json } => {
                     // [impl cli.list.interactive]
                     // [impl cli.list.non-interactive]
-                    if interactive {
-                        crate::tui::run_list(source, filter)
+                    // [impl cli.list.json]
+                    if json || !interactive {
+                        list_battery_packs(&source, filter.as_deref(), json)
                     } else {
-                        // [impl cli.list.query]
-                        // [impl cli.list.filter]
-                        print_battery_pack_list(&source, filter.as_deref())
+                        crate::tui::run_list(source, filter)
                     }
                 }
                 BpCommands::Show {
@@ -317,35 +324,40 @@ pub fn main() -> Result<()> {
                     template,
                     path,
                     define,
+                    json,
                 } => {
-                    let show_opts = crate::tui::ShowOpts {
-                        battery_pack: &battery_pack,
-                        template: template.as_deref(),
-                        path: path.as_deref(),
-                        source,
-                        defines: define.into_iter().collect(),
-                    };
-                    if interactive {
-                        // [impl cli.show.interactive]
-                        // [impl cli.show.template-preview]
-                        crate::tui::run_show(show_opts)
-                    } else if let Some(tmpl) = show_opts.template {
-                        // [impl cli.show.template-preview]
-                        print_template_preview(&crate::template_engine::PreviewOpts {
-                            battery_pack: show_opts.battery_pack,
-                            template: tmpl,
-                            path: show_opts.path,
-                            source: &show_opts.source,
-                            defines: show_opts.defines,
-                        })
-                    } else {
-                        // [impl cli.show.non-interactive]
-                        print_battery_pack_detail(
-                            show_opts.battery_pack,
-                            show_opts.path,
-                            &show_opts.source,
+                    // [impl cli.show.json]
+                    // [impl cli.show.non-interactive]
+                    if json || (!interactive && template.is_none()) {
+                        show_battery_pack(
+                            &battery_pack,
+                            path.as_deref(),
+                            &source,
                             &project_dir,
+                            json,
                         )
+                    } else {
+                        let show_opts = crate::tui::ShowOpts {
+                            battery_pack: &battery_pack,
+                            template: template.as_deref(),
+                            path: path.as_deref(),
+                            source,
+                            defines: define.into_iter().collect(),
+                        };
+                        if interactive {
+                            // [impl cli.show.interactive]
+                            // [impl cli.show.template-preview]
+                            crate::tui::run_show(show_opts)
+                        } else {
+                            // [impl cli.show.template-preview]
+                            print_template_preview(&crate::template_engine::PreviewOpts {
+                                battery_pack: show_opts.battery_pack,
+                                template: show_opts.template.unwrap(),
+                                path: show_opts.path,
+                                source: &show_opts.source,
+                                defines: show_opts.defines,
+                            })
+                        }
                     }
                 }
                 BpCommands::Status { path, json } => {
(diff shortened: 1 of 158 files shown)


src/battery-pack/bphelper-cli/src/commands/mod.rs
```diff
@@ -748,11 +745,11 @@ pub(crate) fn add_battery_pack(
...
-        let manifest_content = std::fs::read_to_string(&manifest_path)
-            .with_context(|| format!("Failed to read {}", manifest_path.display()))?;
-        let spec = bphelper_manifest::parse_battery_pack(&manifest_content)
-            .map_err(|e| anyhow::anyhow!("Failed to parse battery pack '{}': {}", crate_name, e))?;
-        (None, spec)
+        let spec = parse_battery_pack_from_path(&manifest_path)
+            .map_err(|err| anyhow!("Failed to parse battery pack '{}': {}", crate_name, err))?;
+        let version = spec.version.clone(); // !quite sure?
+
+        (Some(version), spec)
```
jlizen: I think we still want to return `None` here, generally you would not want to pin a version when using a local path

src/battery-pack/bphelper-manifest/src/test_support.rs
```diff
@@ -0,0 +1,61 @@
...
+
+    pub fn finalize(&mut self) -> &Path {
+        let member_toml = self
+            .members
+            .iter()
+            .map(|member| format!("\"{member}\""))
+            .collect::<Vec<_>>()
+            .join(",\n");
+
+        let workspace = format!("[workspace]\nresolver \"2\"\nmembers = [\n{member_toml}\n]\n");
```
jlizen: This is not valid TOML, it should be `resolver = "2"`. Are your unit tests passing locally? CI is red earlier due to formatting (we should probably split that out into separate jobs to be less annoying...)

src/battery-pack/bphelper-manifest/src/lib.rs
```diff
@@ -734,155 +700,102 @@ pub fn parse_battery_pack(manifest_str: &str) -> Result<BatteryPackSpec, Error>
...
-
-    // Walk up looking for a workspace.
-    let mut dir = crate_root.to_path_buf();
-    while dir.pop() {
-        if let Ok(specs) = discover_battery_packs(&dir) {
-            return Ok(specs);
-        }
-    }
+/// Runs `cargo metadata` against the given manifest and returns the spec for the matching package.
+pub fn parse_battery_pack_from_path(manifest_path: &Path) -> Result<BatteryPackSpec, Error> {
```
jlizen, line 1002: We should canonicalize the path here, we are mixing relative paths from various inputs but cargo_metadata just does absolute, so our path equality checks will fail.

Let's also make sure we have a unit test guarding this case.

src/battery-pack/bphelper-manifest/src/lib.rs
```diff
@@ -652,74 +632,60 @@ struct RawTemplateSpec {
...
-                    )
-                })
-                .collect()
+    let features = pkg
+        .features
+        .iter()
+        .filter(|(key, value)| {
+            !(optional_dep_names.contains(key.as_str())
+                && value.len() == 1
+                && value[0] == format!("dep:{}", key))
```
jlizen: ++ To this change to strip the auto-gen features.

This made me notice an existing bug. Suppose a bp author specifies:
```
  [features]
  indicators = ["dep:indicatif", "dep:console"]
```

Currently we look for exactly "indicatif" during feature resolution so "dep:indicatif" wouldn't match and thus would error out.  The fix would be to strip the `dep:` prefix in `add_feature_crates` and `validate_features`.

Up to you if you care to scope creep this PR into fixing it, while you're anyway working on the parser, glad to review it. Maybe up for cutting a follow up issue if you don't have time to add that in here?

src/battery-pack/bphelper-manifest/src/lib.rs
```diff
@@ -652,74 +632,60 @@ struct RawTemplateSpec {
...
-    features: Vec<String>,
-    optional: bool,
-}
-
 // ============================================================================
 // Parsing
 // ============================================================================
 
-/// Parse a battery pack's Cargo.toml into a `BatteryPackSpec`.
-pub fn parse_battery_pack(manifest_str: &str) -> Result<BatteryPackSpec, Error> {
```
jlizen, line 676: I'm on board with this refactor to accept the `&Path` directly, but let's label the commit as `refactor(manifest)!` so that it shows up in our changelog as breaking (and we'll bump semver)

src/battery-pack/bphelper-manifest/src/lib.rs
```diff
@@ -33,6 +38,8 @@ pub enum Error {
         #[source]
         source: std::io::Error,
     },
+    #[error("cargo metadata failed: {0}")]
+    Metadata(String),
```
jlizen: ++ to not leaking the cargo_metadata type, but can we keep an opaque version of the error to preserve error chains?

ie:
```
  #[error("cargo metadata failed: {0}")]
  Metadata(#[source] Box<dyn std::error::Error + Send + Sync>),
```

src/battery-pack/bphelper-manifest/src/lib.rs
```diff
@@ -734,155 +700,102 @@ pub fn parse_battery_pack(manifest_str: &str) -> Result<BatteryPackSpec, Error>
...
-    let content = std::fs::read_to_string(&cargo_toml).map_err(|e| Error::Io {
-        path: cargo_toml.display().to_string(),
-        source: e,
-    })?;
-    let spec = parse_battery_pack(&content)?;
-    Ok(vec![spec])
-}
+    // -- find the package whose manifest path matches the requested one --
+    let manifest_utf8 = Utf8Path::from_path(manifest_path)
+        .ok_or_else(|| Error::Metadata("manifest path is not UTF-8".into()))?;
```
jlizen: worth flushing the path here as well, maybe? ie roughly:
```
  .ok_or_else(|| Error::Metadata(
      format!("manifest path is not valid UTF-8: {}", manifest_path.display()).into()
  ))?;
```

src/battery-pack/bphelper-manifest/src/lib.rs
```diff
@@ -734,155 +700,102 @@ pub fn parse_battery_pack(manifest_str: &str) -> Result<BatteryPackSpec, Error>
...
+        .manifest_path(manifest_path)
+        .no_deps()
+        .exec()
+        .map_err(|err| Error::Metadata(err.to_string()))
+}
 
-    Ok(packs)
+/// Discover battery packs reachable from a path.
+///
+/// `path` may be a workspace root or any crate withing a workspace;
```
jlizen:
```suggestion
/// `path` may be a workspace root or any crate within a workspace;
```

src/battery-pack/bphelper-manifest/src/lib.rs
```diff
@@ -652,74 +632,60 @@ struct RawTemplateSpec {
...
     let mut crates = BTreeMap::new();
-    parse_dep_section(&raw.dependencies, DepKind::Normal, &mut crates);
-    parse_dep_section(&raw.dev_dependencies, DepKind::Dev, &mut crates);
-    parse_dep_section(&raw.build_dependencies, DepKind::Build, &mut crates);
+    for dep in &pkg.dependencies {
+        let kind = match dep.kind {
+            DependencyKind::Normal => DepKind::Normal,
+            DependencyKind::Development => DepKind::Dev,
+            DependencyKind::Build => DepKind::Build,
+            _ => continue, // or unreachable!().
```
jlizen: this is fine, let's say `// skip unknown kind fields`

We could add a debug log here maybe

src/battery-pack/bphelper-cli/src/validate/mod.rs
```diff
@@ -149,7 +148,9 @@ pub fn validate(manifest_dir: &str) -> Result<()> {
             );
             (manifest_dir.to_path_buf(), false)
         }
-        Err(e) => return Err(e),
+        Err(e) => {
```
jlizen: surprised that this passes rustfmt?

src/battery-pack/bphelper-manifest/src/lib.rs
```diff
@@ -652,74 +632,60 @@ struct RawTemplateSpec {
...
+            !(optional_dep_names.contains(key.as_str())
+                && value.len() == 1
+                && value[0] == format!("dep:{}", key))
         })
-        .unwrap_or_default();
+        .map(|(key, value)| (key.clone(), value.iter().cloned().collect()))
+        .collect::<BTreeMap<_, BTreeSet<_>>>();
+
+    // -- side-read TOML for [package.metadata.battery-pack].hidden + battery.templates --
+    let (hidden, templates) = read_bp_metadata_tables(&pkg.manifest_path)?;
```
jlizen: > Side-read [package.metadata.battery-pack] and [package.metadata.battery.templates] from TOML directly -- cargo_metadata exposes these as serde_json::Value which loses the original shape.

I don't think we care about that here, no? Since we are anyway not writing the toml back out into user space files. We only are parsing its contents, which shouldn't be lossy except for eg formatting and comments?

But the `[package.metadata.battery-pack].hidden` + `battery.templates` should propagate properly?

IE, seems like we could skip the extra disk read and just use the `cargo_metadata` output directly?

src/battery-pack/bphelper-manifest/src/test_support.rs
gmemuriuki: Ooh, I see. 🤦🤦...

gmemuriuki: Sure, formatting drives me crazy all the time in CI.

Mostly because sth is broke in my editor such that it doesn't run `cargo fmt`automatically on save. Unless I run it manually. Quite sad actually.

src/battery-pack/bphelper-manifest/src/lib.rs
gmemuriuki: 👍 makes sense.

src/battery-pack/bphelper-manifest/src/test_support.rs
jlizen: Could always add a global git pre-commit hook that runs cargo fmt (or non-ops if not in a cargo workspace).

I'd probably fix the editor though ^^
Or change editors if is truly difficult to get it working right.

src/battery-pack/bphelper-manifest/src/lib.rs
gmemuriuki: I don't think it involves a lot of code changes so, I'm gonna do it here.

src/battery-pack/bphelper-manifest/src/lib.rs
gmemuriuki: Awesome. 👍

src/battery-pack/bphelper-manifest/src/lib.rs
gmemuriuki: 🤦‍♀️🤦‍♀️

src/battery-pack/bphelper-manifest/src/lib.rs
gmemuriuki: Yeah, we can do that.

src/battery-pack/bphelper-manifest/src/lib.rs
gmemuriuki: 💯

src/battery-pack/bphelper-manifest/src/lib.rs
gmemuriuki: I did not know that actually. Thank you.

src/battery-pack/bphelper-manifest/src/lib.rs
gmemuriuki: I think it makes total sense to handle all the case such as:
```toml
 [features]
 indicators = ["indicatif/foo", "console/?foo"]
```
don't you think?

src/battery-pack/bphelper-manifest/src/lib.rs
gmemuriuki: Yeah. Totally possible. I'll fix that too.

src/battery-pack/bphelper-manifest/src/lib.rs
jlizen: Yeah I'd agree with that. Ideally we could just use cargo provided resolution rather than hand rolling it, but if that is a larger change, making the hand rolled cover all these cases to start with is good.

src/battery-pack/bphelper-cli/src/validate/mod.rs
```diff
@@ -112,15 +113,13 @@ pub(crate) fn validate_battery_pack_cmd(path: Option<&str>) -> Result<()> {
...
-        .with_context(|| format!("failed to read {}", cargo_toml.display()))?;
 
     let crate_name = manifest_dir
         .file_name()
         .and_then(|s| s.to_str())
         .unwrap_or("unknown");
-    let spec = bphelper_manifest::parse_battery_pack(&content)
-        .map_err(|e| anyhow::anyhow!("failed to parse {}: {e}", cargo_toml.display()))?;
+    let spec = parse_battery_pack_from_path(&cargo_toml)
+        .map_err(|err| anyhow!("failed to parse {}: {}", cargo_toml.display(), err))?;
```
JayanAXHF: I've noticed that a lot of places have a pattern of `map_err(|err| anyhow!("something {}: {}", smth, err))`. Have you considered making that a function to abstract some logic away?

src/battery-pack/bphelper-manifest/src/lib.rs
```diff
@@ -1018,8 +956,108 @@ fn validate_templates_on_disk(
...
+        version = "0.1.0"
+        keywords = ["battery-pack"]
+
+        [dependencies]
+        indicatif = {version = "0.17", optional = true}
+
+        [features]
+        indicators = ["dep:indicatif"]
+        "#;
+
```
JayanAXHF, line 1217: for these, consider using `snapbox` or something similar. I'm not quite on board with toml embedded right into the test code like this.

JayanAXHF (dismissed):

.gitignore
```diff
@@ -2,3 +2,4 @@ book
 target
 WIP.md
 .claude/settings.local.json 
+.claude/
```
jlizen: let's only keep commands and reject all else, ie make sure when we rebase against #134 we take that gitignore

src/battery-pack/bphelper-cli/src/registry/mod.rs
```diff
@@ -247,17 +248,16 @@ pub(crate) fn fetch_bp_spec_from_registry(
...
-        .map_err(|e| anyhow::anyhow!("Failed to parse battery pack '{}': {}", crate_name, e))?;
-
-    // Cache the manifest dynamically for autocomplete
+    // Cache the parsed spec as JSON for autocomplete
     let cache_dir = crate::completions::get_cache_dir();
-    if fs::create_dir_all(&cache_dir).is_ok() {
-        let cache_file = cache_dir.join(format!("{}_spec.toml", crate_name));
-        let _ = fs::write(&cache_file, &manifest_content);
+    if fs::create_dir_all(&cache_dir).is_ok()
+        && let Ok(json) = serde_json::to_string(&spec)
```
jlizen: would be nice to have a debug log in case of deserialize failures

jlizen (commented):
(I think you are still working through feedback because noticed a couple more things)

src/battery-pack/bphelper-cli/src/validate/mod.rs
gmemuriuki: Wait, i don't quite follow. What do you mean?

src/battery-pack/bphelper-manifest/src/lib.rs
gmemuriuki: I see embedded toml everywhere in the codebase. I'm not quite sure whether I should change for those test that I've written, or all the tests.

.gitignore
gmemuriuki: Aah, totally.

src/battery-pack/bphelper-cli/src/registry/mod.rs
gmemuriuki: Sure. 👍

src/battery-pack/bphelper-manifest/src/lib.rs
gmemuriuki: Anyway, I could do the changes as follow-up PR.

src/battery-pack/bphelper-cli/src/validate/mod.rs
jlizen: You can also use `.context()` to help with that:
https://docs.rs/anyhow/latest/anyhow/trait.Context.html

src/battery-pack/bphelper-manifest/src/lib.rs
jlizen: ++ to using snapbox to manage snapshots instead of making point assertions like `contains_key`. Or rather, in addition to, we can keep the point assertion as well.

And it looks like we use `indoc!` to manage these embedded inputs.

This test is pretty much the same as we want, I think:
https://github.com/battery-pack-rs/battery-pack/blob/103d3f139ebf1ffdd4185fa9d31fce0e7bb929eb/src/battery-pack/bphelper-cli/src/merge/tests.rs#L660

src/battery-pack/bphelper-manifest/src/lib.rs
```diff
@@ -1018,8 +956,108 @@ fn validate_templates_on_disk(
 
 #[cfg(test)]
```
jlizen, line 1156: Can we also have a regression test for glob members (`(members = ["crates/*"])`)? That would have broke old parser but should work now.

src/battery-pack/bphelper-manifest/src/lib.rs
```diff
@@ -1751,7 +1804,7 @@ mod tests {
 
     #[test]
     // [verify cli.source.discover] workspace case — member crate discovers siblings
-    fn discover_from_crate_root_finds_workspace() {
+    fn discover_battery_packs_finds_workspace() {
```
jlizen, line 2272: can we also have a negative test that non-battery-pack workspace members are excluded

src/battery-pack/bphelper-cli/src/commands/mod.rs
jlizen: Same concern @gme-muriuki

jlizen (commented):
Good refactors, a few more small things @gme-muriuki

src/battery-pack/bphelper-manifest/src/lib.rs
gmemuriuki: Yeah 👍

src/battery-pack/bphelper-manifest/src/lib.rs
gmemuriuki: Totally possible. I'll do that too.

gmemuriuki:
> Same concern @gme-muriuki 

Sorry, totally slipped my mind

src/battery-pack/bphelper-manifest/src/lib.rs
```diff
@@ -176,6 +181,18 @@ pub struct BatteryPackSpec {
...
 
+/// Extract the crate name from a Cargo feature reference.
+///
+/// - `"foo"`
+/// - `"dep:foo"`
+/// - `"foo/bar"`
+/// - `"foo?/bar"`
+fn crate_from_feature_reference(s: &str) -> &str {
+    let crt = s.strip_prefix("dep:").unwrap_or(s);
+    let crt = crt.split('/').next().unwrap_or(crt);
```
jlizen: This behavior will drop any optional features, ie this:
```
 [dependencies]
  serde = { version = "1", optional = true }

  [features]
  fancy = ["serde/derive"]
```

Resolves to:
```
serde = "1" 
```

instead of:
```
serde = { version = "1", features = ["derive"] }
```

This seems like a regression to me, I think we want the latter. 

I would probably handle this via a small local parser like so:
```
  enum FeatureRef<'a> {
      Dep(&'a str),
      Feature(&'a str),
      DepFeature {
          dep: &'a str,
          feature: &'a str,
          weak: bool,
      },
  }
```

(though maybe @epage has better advice)

jlizen (changes requested):

src/battery-pack/bphelper-manifest/src/lib.rs
gmemuriuki: We could defer to cargo entirely and own as minimum parsing as possible.

gmemuriuki: Thinking of going with the typed-enum direction you suggested, but lifting the parse out of the resolver into manifest-load time -- so BatteryPackSpec.features stores `Vec<FeatureRef>` instead of `Vec<String>`.
Variants would be Dep, Feature, and DepFeature {dep, feature, weak }. 

Validation moves up with it: bad refs become parse errors with location, instead of silent drops at resolve time.

The resolver becomes a small match -- for DepFeature it unions the per-dep feature into the result's feature set, which is the actual fix.

One question I want your take on:
 - I'm planning to treat weak (foo?/bar) and strong (foo/bar) the same in the resolver. The reasoning is we're a recommender, not a runtime resolver -- weak/strong only affects whether cargo auto-activates the dep at runtime, which is the downstream consumer's concern. The weak flag would still be a FeatureRef so a future emitter could mirror it. 

Does that match your mental model, or would you rather drop the weak ref entirely (closer to cargo's runtime behaviour) at resolve time?

For the broader "what about future cargo semantic changes?" worry, I'm thinking of adding an oracle test harness behind an oracle cargo feature: per-fixture x feature combo, assert our resolver names the same direct deps as cargo metadata --features <combo>. Run it with CI. That way if cargo's rules shift, the test fails before users hit it.

Open to alternatives if you've seen this done another way.

cc @epage

jlizen: > Thinking of going with the typed-enum direction you suggested, but lifting the parse out of the resolver into manifest-load time -- so BatteryPackSpec.features stores Vec<FeatureRef> instead of Vec<String>.
Variants would be Dep, Feature, and DepFeature {dep, feature, weak }.

Seems good

> I'm planning to treat weak (foo?/bar) and strong (foo/bar) the same in the resolver. The reasoning is we're a recommender, not a runtime resolver -- weak/strong only affects whether cargo auto-activates the dep at runtime, which is the downstream consumer's concern. The weak flag would still be a FeatureRef so a future emitter could mirror it.

This seems fine to treat them identically for feature propagation, but we should be careful not to activate dependencies if there is only a weak syntax and nothing else is directly activating the feature

> For the broader "what about future cargo semantic changes?" worry, I'm thinking of adding an oracle test harness behind an oracle cargo feature: per-fixture x feature combo, assert our resolver names the same direct deps as cargo metadata --features . Run it with CI. That way if cargo's rules shift, the test fails before users hit it.

An oracle seems nice but we would want to be careful about complexity explosion or flakiness. Probably I'd suggest using a path dependency in fixtures to avoid network usage. And keep the fixture cases relatively few and targeted (weak deps, dep: syntax, feature-of-a-dep, etc).

gmemuriuki: Makes sense 👍. Working on it, will be through mid next week.

src/battery-pack/bphelper-cli/src/commands/mod.rs
```diff
@@ -1430,9 +1428,10 @@ fn pick_crates_interactive(
         match item {
             PickerItem::Feature(feat_name) => {
                 if let Some(members) = bp_spec.features.get(feat_name) {
-                    for c in members {
+                    for fref in members {
+                        let c = fref.dep_name();
```
jlizen: This needs to pass through the new resolver that non-interactive mode now uses. Otherwise, it skips feature-gated features

src/battery-pack/bphelper-cli/src/commands/mod.rs
jlizen: this needs the new resolver

src/battery-pack/bphelper-manifest/src/lib.rs
jlizen: needs new resolver

jlizen (changes requested):
I think we should remove `resolve_all_visible()` as a public API, I think I found all the usages but clearly it is a maintenance footgun.

I also think we are missing some e2e tests. The unit tests are comprehensive but we have holes where this feature resolution gap would have passed through.

Suggested tests:
  - `add --no-default-features -F fancy` preserves per-dep features
  - `add --all-features` also preserves per-dep features
  - Nested refs like `bundle = ["fancy"]` expand correctly

gmemuriuki:
I'm in a bit of a situation here. Should the `parse_battery_pack_from_path` [link](https://github.com/battery-pack-rs/battery-pack/pull/135/changes#diff-35e6c0db2e10491037ae187bd9f472ff12844a8899653f34c7247fc000e061ebR916) validate eagery?

Context: the manifest crate exposes `parse_battery_pack_from_path(&Path) -> Result<BatteryPackSpec, Error>` (runs cargo metadata, returns the spec for the matching package). The spec also has a separate BatteryPackSpec::validate() method that checks three things: the package name ends in -battery-pack, every feature reference points at a real crate, and there are no cycles in local-feature-edge.

Right now `parse_battery_pack_from_path` does NOT call validate(). Callers can use the spec without ever validating it.

I tried fixing this by calling the validate() inside the function. It broke `bphelper-build`  .  That crate calls the same parse function from a build rs(which comes from bphelper-build) ... that runs inside template-scaffolded user crates.  The scaffolded crate isn't a battery pack itself -- its name is bp-validate-default, not *-battery-pack -- so eager-validate panics it with InvalidName.

So there are two different consumers with different contracts:
- bphelper-build -- runs inside scaffolded user crates whose names won't match. Needs lenient parse.
- bphelper-cli

Should I split them into two entry points:
```rust
 /// lenient, what bphelper-build uses
pub fn parse_battery_pack_from_path(path: &Path) -> Result<BatteryPackSpec, Error> {
     ...
}

```

```rust
/// parse + validate
pub fn parse_and_validate_battery_pack(path: &Path) -> Result<BatteryPackSpec, Error> {
    ...
    spec.validate()?;
}

```

JayanAXHF:
> Should I split them into two entry points:

I don't believe that that is necessary. As I understand it to be, parsing and validation are different things. Just as well, I can conjure up cases where you _wouldn't_ need to validate: for example if you knew that the spec in question was for a battery pack. I think it'd be fine if you just put a doc comment in `parse_battery_pack_from_path` warning users about the `validate` call.

gmemuriuki:
> ... I think it'd be fine if you just put a doc comment in parse_battery_pack_from_path warning users about the validate call.

Interesting. This is what I have in the code as of now. If that's the right call, I'll commit to that then. @jlizen What do you think?

src/battery-pack/bphelper-manifest/src/lib.rs
```diff
@@ -337,31 +515,24 @@ impl BatteryPackSpec {
...
-        }
+    /// `ActiveFeatures::All` expands to every declared feature so per-dep activations
+    /// (`serde/derive`, `foo?/bar`, nested refs) reach `resolve_crates` instead of being dropped.
+    /// Hidden crates are always excluded from the result -- this is the surface every
+    /// user-facing caller (sync, status, `--all-features`, bp-managed rewrite) reads from.
+    ///
+    /// [impl format.hidden.effect]
+    pub fn resolve_for_features(&self, active: &ActiveFeatures) -> BTreeMap<String, CrateSpec> {
+        let expanded: Vec<&str> = match active {
+            ActiveFeatures::All => self.features.keys().map(String::as_str).collect(),
```
jlizen, line 534: We have an untested edge case where an optional dependency that has no explicit feature, will be dropped here. This is contrary to cargo since cargo injects the implicit feature for that dependency. (Since we strip those extra features before this method is called). It also is contrary to the older `resolve_all_visible()` handling.

This means our oracle will not match the output.

I don't have a strong opinion about what the right way to handle it is,  but we should have an oracle case covering it and some brief documentation.

src/battery-pack/bphelper-manifest/src/lib.rs
```diff
@@ -141,17 +169,39 @@ pub struct CrateSpec {
...
 
 /// Template metadata for project scaffolding.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct TemplateSpec {
     pub path: String,
     pub description: Option<String>,
 }
 
+/// Active feature selection at the resolver boundary.
```
jlizen, line 178: I suggested just pulling off the bandaid and doing a state format version bump and switching to tagged representation using serde, encapsulating the sentinel here doesn't buy us very much

src/battery-pack/bphelper-manifest/src/lib.rs
```diff
@@ -141,17 +169,39 @@ pub struct CrateSpec {
...
+///
+/// Note: Cargo permits `all` as a feature name (one exists in the `mixed-kind-battery-pack` oracle fixture), so a persisted [`BTreeSet<String>`] containing that literal is ambiguous.
+/// The [`From`] conversion always favors [`ActiveFeatures::All`]; a future tagged on-disk format would remove the ambiguity.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ActiveFeatures {
+    // Every declared feature is active
```
jlizen:
```suggestion
    /// Every declared feature is active
```

jlizen (commented):
> Lenient parse vs two entrypoints

I think a second entrypoint JUST to add name validation is probably overkill. But, we aren't currently calling `validate()` on install or sync, we could add that into `load_installed_bp_spec`. AIUI it mostly would affect `cargo bp add --path` (which would currently allow a non-`*-battery-pack` name).

(And then for `show`/`list`, I don't think we want validation, so lenient is probably the right default?)

We should add a regression test for the CLI capturing that `--path` case and that it is appropriately rejected.

src/battery-pack/bphelper-manifest/src/lib.rs
jlizen: This is scope creep, let's leave it as-is and i'll knock that out as a quick extra bit. Niko is working on a blog post so want to get the protocol bump ahead of time.

jlizen:
I went ahead and addressed my last couple nits, thanks for all the rounds of feedback on this @gme-muriuki ! Big improvement to our internal machinery.

jlizen (approved):

PR #136, opened 2026-05-23
Add optional cross-platform testing job

jlizen:
### Summary

Adds a `cross_platform` opt-in feature that runs the test suite on macOS and Windows via a matrix job. Useful for crates that deal with filesystem paths, platform-specific FFI, or conditional compilation.

The job reuses the existing `rust-build` composite action (stable toolchain, `--all-features`) rather than duplicating the full toolchain/feature matrix from the `build` job. Cross-platform bugs are almost always platform-specific (path separators, line endings, FFI), not toolchain-version-specific, so this keeps CI minutes reasonable.

Inspired by the `test-os` job in [tower-rs/tower-http#683](https://github.com/tower-rs/tower-http/pull/683).

### Testing

The maximalist snapshot includes the new `cross-platform` job and its entry in the `ci-pass` needs list. The minimalist snapshot correctly omits it (opt-in only). `validate` test confirms template rendering is well-formed.


JayanAXHF (approved):
LGTM

PR #138, opened 2026-05-27
fix(ci-battery-pack): make snapshot tests cross-platform for .exe

jlizen:
Quickie - #134 added proper CI tests for windows. But, snapbox has an inbuilt behavior to redact `.exe` suffixes on windows, that doesn't apply on other platforms (weird choice).

Changing it to unconditionally apply the redaction to avoid mismatch between linux vs windows snaps.

The actual user-facing functionality was fine all along.


jlizen:
Fyi @gme-muriuki ... mystery solved

gmemuriuki:
> Fyi @gme-muriuki ... mystery solved

🤣🤣, Awesome... 🤣 Good work 👏

PR #140, opened 2026-05-31
feat(cargo-bp-script): add JSON output for `list` and `show` commands

nikomatsakis:
Extend cargo-bp-script with schema types and runners for `cargo bp list --json` and `cargo bp show --json`, following the same pattern as the existing `status --json` support. Add --json flags to both CLI subcommands and integration tests exercising the binary and the typed runner API.

### DIFF
--- a/src/battery-pack/bphelper-cli/src/commands/mod.rs
+++ b/src/battery-pack/bphelper-cli/src/commands/mod.rs
@@ -181,7 +181,7 @@ pub(crate) enum BpCommands {
         define: Vec<(String, String)>,
 
         /// Emit machine-readable JSON instead of the default text output
-        #[arg(long)]
+        #[arg(long, conflicts_with_all = ["template", "define"])]
         json: bool,
     },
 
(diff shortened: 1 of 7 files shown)


src/battery-pack/bphelper-cli/src/commands/mod.rs
```diff
@@ -316,35 +327,41 @@ pub fn main() -> Result<()> {
...
-                        print_template_preview(&crate::template_engine::PreviewOpts {
-                            battery_pack: show_opts.battery_pack,
-                            template: tmpl,
-                            path: show_opts.path,
-                            source: &show_opts.source,
-                            defines: show_opts.defines,
-                        })
+                    // [impl cli.show.json]
+                    if json {
+                        show_battery_pack_json(&battery_pack, path.as_deref(), &source)
```
jlizen: This ignores any `--template` (and template `--define`) flags, ie the previews. Ideally we would instead output a json object of file names and values (though then it needs to be escaped, which, gross.

It's fine to punt but probably should be a validation error if both are provided.

src/cargo-bp/tests/list_json.rs
```diff
@@ -0,0 +1,132 @@
+//! Integration tests for `cargo bp list --json`.
```
jlizen, line 1: any reason for no snapbox with these? would be nice to see the output directly

jlizen (approved):
Does what it needs to, I'd suggest addressing the comments here since it's quick, but no objection to merging as is

PR #142, opened 2026-05-31
feat: add backend service battery pack

jlizen:
### Summary

*(AI disclosure: I used a LLM (claude Opus 4.8) for most of the literal codegen, but I was in the loop the whole time, and heavily refined approach, verbiage, and manually exercised the application myself both via direct CLI usage and agentic benchmarks. Most of the source content for the templates and skills was human-written as well, from closed-source guides and repositories.)*

Adds `network-service-battery-pack`, a new battery pack that scaffolds an opinionated async network service with structured observability, downstream resilience, and graceful shutdown out of the box.

I considered names like `opinionated-network-service-battery-pack` (too clunky) and `opinionated-server-battery-pack` (too narrow for future async queue processing). Settled on `network-service-battery-pack` since all battery packs are opinionated by nature, and "network service" is broad enough to cover HTTP servers, queue processors, and other networked workloads we plan to add later.

The template generates a complete axum service with:
- Wide-event metrics via metrique (JSON to stdout, or rolling files when `--telemetry-dir` is set)
- Structured tracing with non-blocking appender
- Optional dial9 flight recorder integration (cpu profiling, heap profiling, Tokio telemetry)
- jemalloc/mimalloc allocator selection (target-gated for MSVC), wrapped in dial9's profiling allocator when dial9 is enabled
- Graceful shutdown with drain timeout and shutdown metrics
- Downstream backend: in-memory when no `--downstream-url` is passed (zero dependencies to run), HTTP forwarding with reqwest native retry when one is (and also a nice local test setup that can chain two servers)
- Proper demonstrating of reusing reqwest clients across handlers
- Tower middleware stack (timeout, catch-panic, on-early-drop, telemetry), optional inbound rate limiting via tower-governor, various outher layers
- Criterion benchmarks exercising the handler via `tower::oneshot`
- End-to-end test that spawns two instances (backing store + forwarder) and asserts on telemetry output

I also ship three supporting skills:
- telemetry
- memory-allocator
- service-architecture

I think probably memory-allocator belongs in some symposium best practices skill, but we can leave it here for now.

**Not yet included (future work tracked in TODO.md):** async queue processing template, redis/kafka/postgres downstream modules, raw Tower/hyper instead of axum, TLS & mTLS, gRPC, OTEL/OTLP formatter support with metrique (and probably tracing), kafka.

I initially had a few more knobs (redis downstream, a circuit breaker, an in-memory cache for the backend, etc), but I opted to slim down and move most of those into skills as optional additions. I think we probably will want to decompose things a bit more into sub-templates before we start really getting into the more complex application patterns that involve more tradeoffs. Question for another day.

I was also a bit torn on demonstrating nicer EMF patterns (dimensions, constraining numeric values to being properties, etc). I settled for vanilla JSON since at least it's neutral, and our skill can mention formats. Once we have metrique otel support (pending release, PR landed), I'd feel more comfortable adding a flag that nicely wires up metrics based on format.

**Scope creep during implementation:**
- `bphelper-cli` template engine now runs rustfmt as a post-render step for `.rs` files, and `cargo bp validate` asserts `cargo fmt --check` passes. This eliminated a class of template whitespace bugs where MiniJinja `{%- -%}` trimming interacted poorly with Rust formatting. Templates now use plain `{% %}` tags for Rust files and let rustfmt normalize the output.
- `bphelper-manifest`: fixed a bug where non-optional base dependencies were dropped from resolution when explicit features were active.
- `bphelper-cli` validate: fixed `.cargo/config.toml` patch injection overwriting existing content (now merges tables).
- `bphelper-cli` validate: added `validate_template_with` for exercising multiple placeholder combinations in a single battery pack.
- Bumped test fixture editions from 2021 to 2024.
- Fixed some templates that still generated with `[package.metadata.battery-pack]` instead of using `battery-pack.toml` (which first required teaching `bp-managed` about `battery-pack.toml`)

### Testing

All placeholder combinations compile, pass their own tests, and pass `cargo fmt --check` via `cargo bp validate`. The generated service includes unit tests (metrics, tracing, handler behavior via oneshot) and an e2e test that spawns two real binary instances to validate forwarding and telemetry output. The manifest resolution bug fix has a dedicated regression test.

There is also a "benchmark" that bootstraps the project, interacts with it via CLI and with dial9, and adds some of the "extra stuff" from skills. It shows that it is basically comprehensible by an agent, here's an example:

https://gist.github.com/jlizen/a79cefc5914d9a834063b49e9e361730


src/battery-pack/bphelper-cli/src/registry/mod.rs
```diff
@@ -269,7 +269,40 @@ pub(crate) fn fetch_bp_spec_from_registry(
 
 /// Resolve `bp-managed = true` dependencies in a Cargo.toml string,
 /// returning the rewritten content with concrete versions.
-pub fn resolve_bp_managed_content(content: &str, bp_crate_root: &Path) -> Result<String> {
+/// Parse `(pack short name, features)` from a rendered `battery-pack.toml` state file. Unparsable
+/// content yields no packs; an entry without features defaults to `default`.
+fn active_packs_from_state(state: &str) -> Vec<(String, std::collections::BTreeSet<String>)> {
```
jlizen, line 274: I suspect this overlaps with #140 , feel free to rebase accordingly @nikomatsakis

jlizen:
Per offline discussion:
- figuring out graceful shutdown on windows (that is breaking our tests, correctly, missing a final application log flush)
- moving to `battery-pack-rs/opionated` repo that @nikomatsakis will create

jlizen:
Got the graceful shutdown handling on windows working properly (with e2e test support).

I also started a zulip thread: https://rust-lang.zulipchat.com/#narrow/channel/575623-battery-packs-rs/topic/Opinionated.20battery.20packs/with/599207748

My current thinking is, let's move this into a subdirectory, `opinionated-battery-packs`, since we will probably keep finding edge cases needing fixes to core bp infrastructure, while working on these templates for some time. And coordinating PRs across multiple repos is unpleasant.

nikomatsakis:
@jlizen I wonder about something like `backend-service-battery-pack`. I guess I'm not sure what's the difference between a *network service* and a *backend service*.

jlizen:
Probably a square and rectangle situation, I don't mind renaming it to `backend-service-battery-pack` if we expect there to ultimately be a different full stack-oriented bp (probably there should be).

nikomatsakis:
r=me once the rename goes through, i'm sure we'll evolve this over time; this looks like an awesome start!

nikomatsakis (approved):

PR #144, opened 2026-06-04
`cargo bp add` templates

nikomatsakis:
Two changes: 

(a) `cargo bp add` will show templates now
(b) we track templates that have been "added" to a space so they can appear "checked"
(c) when instantiating with a template, we record the template in the generated project state


nikomatsakis:
I think this is subsumed by #146

PR #145, opened 2026-06-09
feat(ci): improvements to license/vulnerabiltiy scan template

jlizen:
Doing some more cleanup on the CI battery pack (and this repo's own CI):

### Add cargo-audit + issue posting for vulnerabilities

My other projects use [the RustSec-managed cargo-audit action](https://github.com/rustsec/audit-check) for their daily vulnerability scans. This action posts an issue for you showing what your vulnerabilities are (with deduping). I am currently adding nicer support for dumping your dependency trees and otherwise improving the action: https://github.com/rustsec/audit-check/pull/52

Right now we only use `cargo-deny` which has no issue posting support (though is a better-rounded CLI tool overall, it handles license checks, etc). I think the rustsec `cargo-audit` action is a better default as most users that would want to see the issues for a nicer way to read why their audit is failing (with dependency tree, etc). We could also invest in making `cargo-deny` more extensible with issue posting, but my knee-jerk was to contribute to the rustsec side first.

I left in `cargo-deny` though scoped to licenses, bans, and sources (ie everything but advisories). It is now opt-in rather than opt-out... I could go either way about it, feedback welcome.

### Add standalone dependency-policy and audit templates

Previously the vulnerability scan was only in `full`, not sure why. Now both have their own directed template.

### Apply to repo

Applied both policies to the repo, which previously did not have vulnerability or policy enforcement. We were mostly clean, except for bringing in a sort of weird license when compiling wasm via `reqwest`, which I wasn't motivated to chase down.

### Testing

I ran the individual commands locally but have not figured out a local github action runner testing workflow that I like. I'll keep an eye on the approval flows on this PR too.


battery-packs/ci-battery-pack/templates/dependency-policy/deny.toml
```diff
@@ -0,0 +1,43 @@
+# https://embarkstudios.github.io/cargo-deny/
+
+[licenses]
+allow = [
```
jlizen, line 4: Slightly more permissive than before, I switched to using `cargo` as a reference rather than `smithy-rs`.

.github/workflows/dependency-policy.yml
```diff
@@ -0,0 +1,32 @@
...
+
+on:
+  push:
+    paths:
+      - ".github/workflows/dependency-policy.yml"
+      - "deny.toml"
+      - "**/Cargo.toml"
+      - "**/Cargo.lock"
+  pull_request:
+    paths:
```
epage: Don't path filters make PRs unmergeable if you make these jobs blocking (which they should be)

jlizen: Thanks, I'll just remove the path filter, `cargo-deny` is quick

nikomatsakis (approved):

Issue #104, opened 2026-04-19
[feat] Placeholder selection in TUI preview


jlizen:
Template preview in the TUI always renders with default placeholder values. For templates with boolean flags or select fields (like ci-battery-pack), there's no way to preview different configurations.

### Proposed

When previewing a template that has placeholders, show a form or toggle UI to adjust placeholder values and re-render the preview. This would let users see exactly what files they'll get before generating.

### Context

From #101 , ci-battery-pack has 8 placeholders that significantly change the output.`ci_platform` can strip all `.github/` files, each bool flag adds/removes files and CI jobs. Preview currently only shows the minimalist default output.

jlizen:
Suggestion from #131 

If we add `cargo bp list ci -t full` support, that could jump into TUI with a view that lets you populate each placeholder. And then you can render preview from that.

A simpler "v0" would also just be to have the existing interactive input via CLI, that then jumps to the materialized template in tui.

Similarly, from the tui "show"/"add" view, we could have a keybind that takes you to the same interactive picker, which then can materialize the template with whatever you select.

PR #146, opened 2026-06-09
Add UI 2.0

nikomatsakis:
Revamp the `cargo bp add` TUI. It now looks like this:

```
                                                              ci-battery-pack v0.1.5

┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│Features:                                                                                                                                        │
│> [ ] ✦ benchmarks [criterion]                                                                                                                   │
│  [ ] ✦ fuzzing [arbitrary, libfuzzer-sys]                                                                                                       │
│  [ ] ✦ xtask [xflags, xshell]                                                                                                                   │
│                                                                                                                                                 │
│Dependencies:                                                                                                                                    │
│  [ ] arbitrary (1, features: derive)                                                                                                            │
│  [ ] criterion (0.8, features: html_reports)                                                                                                    │
│  [ ] libfuzzer-sys (0.4)                                                                                                                        │
│  [ ] xflags (0.3)                                                                                                                               │
│  [ ] xshell (0.2)                                                                                                                               │
│                                                                                                                                                 │
│Actions:                                                                                                                                         │
│  [ ] Add `benchmarks` template — Criterion bench scaffold + Bencher CI                                                                          │
│  [ ] Add `binary-release` template — Cross-platform binary builds for GitHub Releases + cargo-binstall                                          │
│  [ ] Add `clippy-sarif` template — Clippy with GitHub PR annotations via SARIF                                                                  │
│  [ ] Add `full` template — Full CI setup with optional benchmarks, fuzzing, mdbook, spellcheck, and xtask                                       │
│  [ ] Add `fuzzing` template — cargo-fuzz scaffold + CI workflows                                                                                │
│  [ ] Add `mdbook` template — mdBook scaffold + GitHub Pages deployment                                                                          │
│  [ ] Add `mutation-testing` template — Mutation testing with cargo-mutants                                                                      │
│  [ ] Add `spellcheck` template — crate-ci/typos config + CI workflow                                                                            │
│  [ ] Add `stress-test` template — nextest stress test workflow                                                                                  │
│  [ ] Add `trusted-publishing` template — release-plz with OIDC trusted publishing                                                               │
│  [ ] Add `xtask` template — cargo-xtask scaffold with codegen --check                                                                           │
│                                                                                                                                                 │
│                                                                                                                                                 │
│                                                                                                                                                 │
│                                                                                                                                                 │
│                                                                                                                                                 │
│                                                                                                                                                 │
│                                                                                                                                                 │
│                                                                                                                                                 │
│                                                                                                                                                 │
│                                                                                                                                                 │
│                                                                                                                                                 │
│                                                                                                                                                 │
│                                                                                                                                                 │
│                                                                                                                                                 │
└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
↑↓/jk Navigate | Space Toggle | a Toggle section | p Preview | Enter Confirm | Esc Cancel
```

You can press `p` on the "actions" to view a preview. This PR introduces a distinct crate for the widget and then uses it within cargo-bp.

### DIFF
--- a/src/battery-pack/bphelper-cli/src/commands.rs
+++ b/src/battery-pack/bphelper-cli/src/commands.rs
@@ -13,9 +13,9 @@ use std::path::{Path, PathBuf};
 use crate::manifest::{
     add_dep_to_table, dep_kind_section, find_installed_bp_names, find_user_manifest,
     find_workspace_manifest, read_active_features_for_project, read_active_features_from_state,
-    read_managed_deps_for_project, remove_battery_pack_state_entry, remove_deps_by_kind,
-    should_upgrade_version, sync_dep_in_table, write_battery_pack_state, write_deps_by_kind,
-    write_workspace_refs_by_kind,
+    read_applied_templates_from_state, read_managed_deps_for_project, record_applied_template,
+    remove_battery_pack_state_entry, remove_deps_by_kind, should_upgrade_version,
+    sync_dep_in_table, write_battery_pack_state, write_deps_by_kind, write_workspace_refs_by_kind,
 };
 use crate::registry::{
     CrateSource, InstalledPack, TemplateConfig, fetch_battery_pack_detail,
@@ -695,6 +695,10 @@ fn add_template(opts: AddTemplateOpts<'_>) -> Result<()> {
     };
     let results = crate::merge::apply_rendered_files(&files, &apply_opts)?;
     crate::merge::print_summary(&results);
+
+    // Record the applied template in battery-pack.toml.
+    let user_manifest_path = find_user_manifest(opts.project_dir)?;
+    record_applied_template(&user_manifest_path, &crate_name, opts.template)?;
 
     // Print post-merge hints if the template defines any.
     if !hints.is_empty() {
@@ -2344,12 +2348,16 @@ fn build_status_report(
                 bphelper_manifest::ActiveFeatures::Subset(set) => set.iter().cloned().collect(),
             };
 
+            let applied_templates =
+                read_applied_templates_from_state(&user_manifest_path, &pack.spec.name);
+
             cargo_bp_script::InstalledPackStatus::new(
                 &pack.short_name,
                 &pack.spec.name,
                 &pack.version,
             )
             .with_active_features(feature_strings)
+            .with_applied_templates(applied_templates)
             .with_warnings(warnings)
         })
         .collect();
--- a/src/battery-pack/bphelper-cli/src/commands.rs
+++ b/src/battery-pack/bphelper-cli/src/commands.rs
@@ -1388,35 +1388,27 @@ pub(crate) struct PickerResult {
     pub selected_templates: Vec<String>,
 }
 
-/// An item in the picker — either a feature, an individual crate, or a template.
-enum PickerItem {
-    Feature(String),  // feature name
-    Crate(String),    // crate name
-    Template(String), // template name
-}
-
 /// Show an interactive multi-select picker for choosing which crates to install.
 ///
-/// Features are listed first, then individual crates. `pre_selected` contains
-/// crate names already present in the project (for edit mode); when empty,
-/// the pack's default feature set is used for initial selection.
+/// Features are listed first, then individual crates, then actions (templates).
+/// `pre_selected` contains crate names already present in the project (for edit
+/// mode); when empty, the pack's default feature set is used for initial selection.
 ///
 /// Returns `None` if the user cancels.
 fn pick_crates_interactive(
     bp_spec: &bphelper_manifest::BatteryPackSpec,
     pre_selected: &BTreeSet<String>,
 ) -> Result<Option<PickerResult>> {
-    use console::style;
-    use dialoguer::MultiSelect;
+    use sectioned_picker::{PickerOutcome, Section, SectionItem, run_picker};
 
-    // Collect non-default features with their member crates
+    // Collect non-default features with their member crates.
     let features: Vec<(&String, &BTreeSet<bphelper_manifest::FeatureRef>)> = bp_spec
         .features
         .iter()
         .filter(|(name, _)| name.as_str() != "default")
         .collect();
 
-    // Collect all visible crates
+    // Collect all visible crates.
     let visible_crates: Vec<(&String, &bphelper_manifest::CrateSpec)> = bp_spec
         .crates
         .iter()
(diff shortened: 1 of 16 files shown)


jlizen:
Haven't yet looked at the impl, but initial interface feedback:

The new screen feels great. Only tweak might be, maybe use `-` instead of `[ ]` for the actions, so the user doesn't hit return expecting to select (rather than exist the tui and apply).

Beyond that I hit some paper cuts while testing that are largely pre-existing. Glad to deal with them in a follow up issue if you don't feel like it, let me know:

The "merge into template" flow does not seem to accept placeholders selection, both from the TUI or directly via CLI with `cargo bp add ci -t full`.

We don't seem to respect `-N` on `cargo bp add ci -N` (should be a helpful error due to ambiguity).

This also draws attention to the same unfortunate gap of tui preview - no placeholder support (#104 ).

nikomatsakis:
Regarding the actions: you can select multiple. If the user hits enter with *nothing* selected, it just auto-selects the current item (whatever it is). Does that influence your take?

jlizen:
Yeah, I missed that. I would have expected enter to be the selection but I guess not?

Perhaps not the most intuitive UI but if it is consistent with how dependency selection works, seems fine to match.

nikomatsakis:
@jlizen Initially I was going to make them "non-checkmarks" but then I thought people might want the option to apply more than one template.

jlizen:
Seems fine since this is leading into the merge UI right?

nikomatsakis:
@jlizen yes

src/battery-pack/bphelper-cli/src/commands.rs
```diff
@@ -1341,119 +1381,221 @@ fn pick_crates_interactive(
...
+        && has_templates
+    {
+        let battery_pack = ctx.battery_pack.clone();
+        let path = ctx.path.clone();
+        let source = ctx.source.clone();
+        let tmpl_paths = template_paths.clone();
+        let tmpl_names = template_names.clone();
+
+        // Determine the section index for Actions (depends on whether features exist).
+        let actions_section_idx = if has_features { 2 } else { 1 };
```
jlizen, line 1525: This seems fragile.

src/sectioned-picker/src/state.rs
```diff
@@ -0,0 +1,259 @@
...
+            .iter()
+            .any(|e| matches!(e, Entry::Item { checked: true, .. }))
+    }
+
+    /// Extract checked state grouped by section (matching original input order).
+    ///
+    /// Consumes the state and returns a `Vec<Vec<bool>>` where each inner vec
+    /// corresponds to a section in the original input order.
+    pub fn into_results(&mut self) -> Vec<Vec<bool>> {
+        let entries = std::mem::take(&mut self.entries);
```
jlizen, line 133: nit: I don't like this take + the `&mut self`, I'd rather see a singular termination point with `self` input.

src/battery-pack/bphelper-cli/src/tui.rs
```diff
@@ -1573,6 +1573,86 @@ fn highlight_preview(files: &[crate::template_engine::RenderedFile]) -> Text<'st
...
+) {
+    let line_count = content.lines.len() as u16;
+    let mut scroll: u16 = 0;
+
+    loop {
+        let _ = terminal.draw(|frame| {
+            render_standalone_preview(frame, title, &content, scroll);
+        });
+
+        if event::poll(Duration::from_millis(100)).unwrap_or(false)
```
jlizen, line 1590: I would prefer to break out of the loop on error to avoid hanging on repeated terminal errors.

src/battery-pack/bphelper-cli/src/tui.rs
```diff
@@ -1573,6 +1573,86 @@ fn highlight_preview(files: &[crate::template_engine::RenderedFile]) -> Text<'st
     Text::from(lines)
 }
 
+/// Show a scrollable syntax-highlighted preview. Blocks until the user presses Esc.
+pub(crate) fn show_preview(
+    terminal: &mut ratatui::DefaultTerminal,
+    title: &str,
+    content: Text<'static>,
+) {
+    let line_count = content.lines.len() as u16;
```
jlizen, line 1582: We would rather truncate than support giant previews?

jlizen (approved):
Looks good, some non-blocking things. I think we should fix the section ordering fragility and the tui hang risk, but follow up is fine.

src/sectioned-picker/src/lib.rs
```diff
@@ -0,0 +1,210 @@
...
+                            terminal,
+                        };
+                        (actions[idx].handler)(&mut ctx);
+                    }
+                }
+                _ => {}
+            }
+        }
+    }
+}
```
JayanAXHF, line 210: I'd say to have this extracted into an "handle_key_event" function.

src/sectioned-picker/src/lib.rs
```diff
@@ -0,0 +1,210 @@
...
+
+    loop {
+        terminal.draw(|frame| render_picker(frame, title, state, &actions))?;
+
+        if event::poll(Duration::from_millis(100))?
+            && let Event::Key(key) = event::read()?
+        {
+            if key.kind != KeyEventKind::Press {
+                continue;
+            }
```
JayanAXHF, line 175: we should probably upgrade to crossterm `0.29`. that gives us the `.is_press()` utility works better with ratatui `0.30`

src/sectioned-picker/src/lib.rs
```diff
@@ -0,0 +1,210 @@
...
+fn run_picker_loop(
+    terminal: &mut ratatui::DefaultTerminal,
+    title: &str,
+    state: &mut PickerState,
+    mut actions: Vec<PickerAction<'_>>,
+) -> anyhow::Result<PickerOutcome> {
+    let action_keys: Vec<char> = actions.iter().map(|a| a.key).collect();
+
+    loop {
+        terminal.draw(|frame| render_picker(frame, title, state, &actions))?;
```
JayanAXHF, line 168: There's a `ratatui::run()` function which can remove calls to `init()` and `restore()`

src/sectioned-picker/src/lib.rs
```diff
@@ -0,0 +1,210 @@
...
+    sections: Vec<Section>,
+    actions: Vec<PickerAction<'_>>,
+) -> anyhow::Result<PickerOutcome> {
+    let mut state = PickerState::new(sections);
+
+    if state.is_empty() {
+        return Ok(PickerOutcome::Confirmed(Vec::new()));
+    }
+
+    let mut terminal = ratatui::init();
```
JayanAXHF, line 153: Have we considered an inline TUI? I think it would look cleaner

JayanAXHF (commented):
Did an initial pass over the sectioned picker. I'll do a few more passes for the other things as time and exam studies permit me.

src/battery-pack/bphelper-cli/src/tui.rs
JayanAXHF: I mean should probably early exit as `terminal.draw()` errors its almost always a fatal error.

src/battery-pack/bphelper-cli/src/tui.rs
```diff
@@ -1573,6 +1573,86 @@ fn highlight_preview(files: &[crate::template_engine::RenderedFile]) -> Text<'st
...
+            Style::default().fg(Color::Cyan).bold(),
+        )))
+        .centered(),
+        header,
+    );
+
+    let preview = Paragraph::new(content.clone())
+        .block(
+            Block::default()
+                .borders(Borders::BOTTOM)
```
JayanAXHF, line 1643:
```suggestion
            Block::default()
                .border_type(BorderType::Rounded)
                .borders(Borders::BOTTOM)
```
Rounded borders would look cleaner IMO

JayanAXHF (approved):
Are `sectioned_picker` and the TUI in `bphelper-cli` different TUIs?

src/battery-pack/bphelper-cli/src/tui.rs
JayanAXHF: Another thing is that if we hang, the user can't `ctrl-c` either since we capture all input. They'd have to `kill` the process, and then `reset` the terminal and get out of `AlternateScreen`, which is really annoying

JayanAXHF:
image
I did some tinkering @nikomatsakis with an inline design for the tui. i also moved the version title to the footer to condense it. What do you think? I liked how the picker looked so i didn't change that

nikomatsakis:
@JayanAXHF that looks nice! Did you want to push that somewhere or open another PR?

JayanAXHF:
If you push all the changes you want to here, i can possibly stack my PR on this. Or I can wait for this to land and then create a PR, both work for me.

jlizen:
> inline TUI

I like it more too! The full screen context switch always rubbed me the wrong way, it doesn't feel very cargo-y.

IMHO let's ship this PR close to as is, with selected tweaks if @nikomatsakis is so inclined, and then cut over as a follow-on. This one already has been open for quite some time :)

JayanAXHF:
What changes are left before we can cut a follow up? Other than the merge conflicts i mean

jlizen:
I'll fix merge conflicts and merge, you can decide which of the open comments you agree with for your PR

PR #150, opened 2026-07-02
feat(state)!: bump format to v2, persist --all-features without a sentinel string

jlizen:
### Summary

The `battery-pack.toml` state file stored active features as a flat string array (`features = ["default", "indicators"]`), with `"all"` as a magic sentinel for `--all-features`. Since Cargo permits `all` as a literal feature name, this was ambiguous.

This switches to typed handling for `--all-features`

```toml
  [[battery-pack]]
  name = "cli"
  all-features = true

  [[battery-pack]]
  name = "error"
  features = ["default", "indicators"]

```

The read path transparently migrates v1 files on first access (using the existing `From<&BTreeSet<String>>` impl). The write path always emits v2. Older `cargo-bp` versions encountering a v2 file will get a clean "please upgrade" error from the existing version gate.

### DIFF
--- a/src/battery-pack/bphelper-cli/src/commands/mod.rs
+++ b/src/battery-pack/bphelper-cli/src/commands/mod.rs
@@ -781,7 +781,10 @@ pub(crate) fn add_battery_pack(
         && specific_crates.is_empty()
     {
         if let Some(existing) = read_active_features_from_state(&user_manifest_path, &crate_name) {
-            let mut merged: Vec<String> = existing.into_iter().collect();
+            let mut merged: Vec<String> = match existing {
+                bphelper_manifest::ActiveFeatures::All => vec!["all".to_string()],
+                bphelper_manifest::ActiveFeatures::Subset(set) => set.into_iter().collect(),
+            };
             for f in with_features {
                 if !merged.contains(f) {
                     merged.push(f.clone());
@@ -937,7 +940,7 @@ pub(crate) fn add_battery_pack(
     write_battery_pack_state(
         &user_manifest_path,
         &crate_name,
-        &active_features,
+        &(&active_features).into(),
         &crates_to_sync,
     )?;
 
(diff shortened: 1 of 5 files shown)


src/battery-pack/bphelper-cli/src/commands/mod.rs
```diff
@@ -781,7 +781,10 @@ pub(crate) fn add_battery_pack(
         && specific_crates.is_empty()
     {
         if let Some(existing) = read_active_features_from_state(&user_manifest_path, &crate_name) {
-            let mut merged: Vec<String> = existing.into_iter().collect();
+            let mut merged: Vec<String> = match existing {
+                bphelper_manifest::ActiveFeatures::All => vec!["all".to_string()],
```
nikomatsakis: this...looks suspicious?

src/battery-pack/bphelper-cli/src/manifest/mod.rs
```diff
@@ -393,15 +393,68 @@ impl Default for BatteryPackStateFile {
...
-    #[serde(default = "default_feature_set")]
+    #[serde(rename = "all-features", default, skip_serializing_if = "is_false")]
+    all_features: bool,
+    #[serde(default = "default_feature_set", skip_serializing_if = "skip_features")]
     features: BTreeSet<String>,
     #[serde(rename = "managed-deps", default)]
     managed_deps: Vec<ManagedDepEntry>,
 }
 
+fn is_false(v: &bool) -> bool {
```
jlizen, line 413: eww, but needed for serde annotations

src/battery-pack/bphelper-cli/src/commands/mod.rs
jlizen: Agreed, will fix

nikomatsakis (approved):
This looks good

PR #156, opened 2026-07-04
style(picker): improve the picker UI and make it inline

JayanAXHF:
image

1. the picker is now inline, except if the calculated height exceeds the terminal height. In that case, it switches to normal fullscreen TUI mode
2. The title has moved to the left of the footer. 
3. Moved some borders for a cleaner look

@nikomatsakis @jlizen what do you think?

### DIFF
--- a/src/sectioned-picker/src/lib.rs
+++ b/src/sectioned-picker/src/lib.rs
@@ -67,6 +67,13 @@ mod state;
 mod tests;
 
 use crossterm::event::{self, Event, KeyCode, KeyEventKind, KeyModifiers};
+use ratatui::{
+    TerminalOptions, Viewport,
+    crossterm::{
+        ExecutableCommand,
+        terminal::{EnterAlternateScreen, LeaveAlternateScreen, disable_raw_mode, enable_raw_mode},
+    },
+};
 use std::time::Duration;
 
 pub use render::render_picker;
@@ -144,15 +151,36 @@ pub fn run_picker(
     sections: Vec<Section>,
     actions: Vec<PickerAction<'_>>,
 ) -> anyhow::Result<PickerOutcome> {
+    let height: u16 = sections
+        .iter()
+        .map(|sec| sec.items.len() as u16 + 2)
+        .sum::<u16>()
+        + 2;
     let mut state = PickerState::new(sections);
-
     if state.is_empty() {
         return Ok(PickerOutcome::Confirmed(Vec::new()));
     }
 
-    let mut terminal = ratatui::init();
+    let mut is_fullscreen = false;
+    let viewport = if let Ok((_, r)) = ratatui::crossterm::terminal::size()
+        && r > height
+    {
+        Viewport::Inline(height)
+    } else {
+        is_fullscreen = true;
+        Viewport::Fullscreen
+    };
+
+    let mut terminal = ratatui::init_with_options(TerminalOptions { viewport });
+    enable_raw_mode()?;
+    if is_fullscreen {
+        terminal.backend_mut().execute(EnterAlternateScreen)?;
+    }
     let result = run_picker_loop(&mut terminal, title, &mut state, actions);
-    ratatui::restore();
+    disable_raw_mode()?;
+    if is_fullscreen {
+        terminal.backend_mut().execute(LeaveAlternateScreen)?;
+    }
     result
 }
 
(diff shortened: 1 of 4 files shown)


src/sectioned-picker/src/lib.rs
```diff
@@ -144,15 +151,36 @@ pub fn run_picker(
...
+        && r > height
+    {
+        Viewport::Inline(height)
+    } else {
+        is_fullscreen = true;
+        Viewport::Fullscreen
+    };
+
+    let mut terminal = ratatui::init_with_options(TerminalOptions { viewport });
+    enable_raw_mode()?;
```
jlizen, line 175: I think this is redundant? I'd prefer `try_init_with_options` + cleanup guard

src/sectioned-picker/src/render.rs
```diff
@@ -22,29 +25,13 @@ pub fn render_picker(
...
-        Constraint::Length(2),
-        Constraint::Fill(1),
-        Constraint::Length(1),
-    ])
-    .areas(area);
+    let [main_area, footer_area] =
+        Layout::vertical([Constraint::Fill(1), Constraint::Length(1)]).areas(area);
 
     // Update visible height from actual terminal geometry.
     state.visible_height = main_area.height.saturating_sub(2) as usize;
```
jlizen, line 32: I think this should be `saturating_sub(1)` now?

src/sectioned-picker/src/lib.rs
```diff
@@ -144,15 +151,36 @@ pub fn run_picker(
...
+        + 2;
     let mut state = PickerState::new(sections);
-
     if state.is_empty() {
         return Ok(PickerOutcome::Confirmed(Vec::new()));
     }
 
-    let mut terminal = ratatui::init();
+    let mut is_fullscreen = false;
+    let viewport = if let Ok((_, r)) = ratatui::crossterm::terminal::size()
```
jlizen, line 165: This would constrain previews to the inline terminal as well. Is that the experience we want? I would have assumed we want them fullscreen?

My suggestion would be, rather than giving callers `&mut DefaultTerminal`, to instead let actions return data or a view enum, and let the picker manage its own transition.

src/sectioned-picker/src/lib.rs
jlizen, line 3: These docs are stale

jlizen (commented):
I love-love-love the new look, nice job!

Re:
> switches to fullscreen if exceeds terminal height

Is this a common pattern? I don't have a great alternative to suggest, but it seems a bit odd to have disjointed UI behavior that feels non-deterministic from customer perspective. I guess as long as their subsequent keypresses to interact with both views are identical, it's probably fine?

Anyway I'm fine to merge it as is, just wondering out loud.

src/sectioned-picker/src/render.rs
JayanAXHF: Probably, i'll confirm tomorrow after the Hindi exam

src/sectioned-picker/src/lib.rs
JayanAXHF: Oops, my bad, will fix tomorrow

src/sectioned-picker/src/lib.rs
JayanAXHF: I don't get what you mean. `init_with_opts` calls `try_init_with_options` and just `.expect()`s the result. The reason we check for fullscreen is that if the viewport height is more than terminal height, weird stuff starts to happen.

src/sectioned-picker/src/lib.rs
JayanAXHF: Could you clarify? i'm a bit lost here trying to understand what you mean

JayanAXHF:
> but it seems a bit odd to have disjointed UI behavior that feels non-deterministic from customer perspective. I guess as long as their subsequent keypresses to interact with both views are identical, it's probably fine?

It's like `saturating_add`, as the number of list items increases, so does viewport height. But we can't increase after the viewport height reacher the terminal height as that causes weird stuff to happen. So instead, we lock it in fullscreen. Visually, it'd be almost the same, as a `Viewport::Inline(t_height)` is virtually identical to `Viewport::Fullscreen`, so we switch to it and go to alternate screen to not clutter the user's terminal scrollback. Other than that, all interactions are unaffected as this does not do anything anywhere else in the app.

nikomatsakis:
I like it. I tried it out. It felt quite natural!

PR #157, opened 2026-07-06
categories, exclusive picks, and awesome embedded bp

nikomatsakis:
Proposes extending battery pack metadata with categories that group features (and templates) and express pick constraints (at-most-one vs any). Covers UI behavior, template integration, category-linked placeholders, and impact on existing packs.


md/rfds/categories-and-alternatives.md
```diff
@@ -0,0 +1,685 @@
...
+```
+
+These options are manually kept in sync with the `allocator` category's
+features. A new placeholder type `"category"` derives its options directly
+from a category definition:
+
+```toml
+[placeholders.allocator]
+type = "category"
+category = "allocator"
```
jlizen: This seems nice

md/rfds/categories-and-alternatives.md
```diff
@@ -0,0 +1,685 @@
...
+
+## UI: How `cargo bp add embedded` looks
+
+### Interactive (TUI)
+
+The picker renders categories as expandable sections with radio-button (●/○)
+or checkbox ([x]/[ ]) semantics based on the `pick` constraint:
+
+```
+  embedded-battery-pack v0.1.0
```
jlizen: Interface LGTM

md/rfds/categories-and-alternatives.md
```diff
@@ -0,0 +1,685 @@
...
+  ─────────────────────────────────────────────────
+  Space: toggle  ←/→: collapse/expand  Enter: confirm  q: cancel  p: preview
+```
+
+UX details:
+
+- **`at-most-one` categories** use radio-button rendering (●/○). Selecting one
+  deselects any other in the same category. However, it's possible to arrive
+  at a state with multiple selections (e.g., the user previously ran
+  `cargo add` manually). In that case the picker shows the current state
```
jlizen: We should display the conflict more prominently in this case and probably require an additional confirmation

md/rfds/categories-and-alternatives.md
```diff
@@ -0,0 +1,685 @@
...
+### `ci-battery-pack`
+
+This pack is template-heavy. Its 13 templates are currently a flat list, but
+several are alternatives to each other or belong to natural groupings:
+
+```toml
+[package.metadata.battery-pack.categories.release]
+title = "Release Strategy"
+description = "How your crate or binary gets published"
+pick = "at-most-one"
```
jlizen: Many projects have both trusted publishing + binary release, including battery-pack. Good to have the `any` grouping though.

md/rfds/categories-and-alternatives.md
```diff
@@ -0,0 +1,685 @@
...
+But the source of truth is the category definition, not a duplicated options
+list. Adding a new feature to the category automatically makes it available as
+a template option — no `bp-template.toml` edit required.
+
+### Features without a category
+
+Features (and templates) that have no `category` annotation continue to work
+exactly as today — they appear as additive checkboxes in the picker.
+
+### Nesting: sub-categories
```
jlizen: Is the idea here is, you can have a narrow parent, and then still allow multiple in child? Or is that intended to be handled by composing battery packs?

It does seem like we will need the narrow parent + fanned-out child... We don't want to force a new battery pack anytime there is any sort of composition like this...

But the UI is tricky to get right. I suggest something, only display selections within a category, though track last selected children if flipping between categories.

I also think we will want to have for templates, the ability to select multiple categories.

```
[package.metadata.battery-pack.categories.platform]
title = "CI platform"
pick = "at-most-one"

[package.metadata.battery-pack.categories."platform.github"]
title = "Github"
pick = "any"


[package.metadata.battery-pack.categories."platform.gitlab"]
title = "GitLab"
pick = "any"

[package.metadata.battery.templates.publish]
path = "..."
description = "publish"
categories = ["platform.github", "platform.gitlab"]

[package.metadata.battery.templates.check]
path = "..."
description = "check"
categories = ["platform.github"]
```

which makes:

```

    ◉ GitHub Actions
        [x] publish
        [ ] check
    ○ GitLab CI
        [ ] publish
  ```
─────────────────────────────────────────────────
```

jlizen (approved):
Looks good

md/rfds/categories-and-alternatives.md
```diff
@@ -0,0 +1,685 @@
...
+mutation-testing = { path = "...", description = "Mutation testing with cargo-mutants", category = "quality" }
+spellcheck = { path = "...", description = "crate-ci/typos config + CI workflow", category = "quality" }
+clippy-sarif = { path = "...", description = "Clippy with GitHub PR annotations via SARIF", category = "quality" }
+security-scanning = { path = "...", description = "RustSec audit workflow", category = "quality" }
+
+# Docs
+mdbook = { path = "...", description = "mdBook scaffold + GitHub Pages deployment", category = "docs" }
+```
+
+`trusted-publishing` and `binary-release` are alternative release strategies
```
jlizen: No, they are complementary. Both this repository and dial9 have both binary release and crates.io trusted publishing

md/rfds/categories-and-alternatives.md
```diff
@@ -0,0 +1,685 @@
...
+   descriptions (only crates do, via their crates.io description). The
+   `description` field in `features.<name>` is new. Should we also support it
+   for non-categorized features? Seems useful for the picker UI generally.
+
+3. **Cross-pack exclusive categories.** If `embedded-battery-pack` and
+   `stm32-battery-pack` both declare a `hal` category, should the exclusivity
+   constraint span both? Probably not (they're different packs, the user
+   wouldn't install both for the same project). But worth considering.
+
+4. **"None" option.** For `at-most-one` categories, should we auto-generate a
```
jlizen: And then we just add a `(required)` if we have `at least one` at some point? Seems fine

md/rfds/categories-and-alternatives.md
```diff
@@ -0,0 +1,685 @@
...
+1. **Interaction with Cargo globals (future).** If/when Cargo gets
+   mutually-exclusive globals, should battery packs automatically generate the
+   corresponding `[globals]` entries? Probably yes, but this is future work.
+
+2. **Feature descriptions.** Today battery pack features don't have
+   descriptions (only crates do, via their crates.io description). The
+   `description` field in `features.<name>` is new. Should we also support it
+   for non-categorized features? Seems useful for the picker UI generally.
+
+3. **Cross-pack exclusive categories.** If `embedded-battery-pack` and
```
jlizen: I don't think we want by default to have category names collide, we could have opt-in syntax for that if it came up.

md/rfds/categories-and-alternatives.md
```diff
@@ -0,0 +1,685 @@
...
+8. **Author an `embedded-battery-pack`** — dogfood the feature with a real
+   battery pack.
+
+## Open questions
+
+1. **Interaction with Cargo globals (future).** If/when Cargo gets
+   mutually-exclusive globals, should battery packs automatically generate the
+   corresponding `[globals]` entries? Probably yes, but this is future work.
+
+2. **Feature descriptions.** Today battery pack features don't have
```
jlizen: Yes, I would include it (not only for non-categorized)

md/rfds/categories-and-alternatives.md
```diff
@@ -0,0 +1,685 @@
...
+description = "Convert panics to 500 responses"
+```
+
+The picker would render allocators as radio buttons (picking one deselects the
+other), and middleware layers as a checkbox group under a clear heading. Today
+these are all in a flat list where the structure is invisible.
+
+### `ci-battery-pack`
+
+This pack is template-heavy. Its 13 templates are currently a flat list, but
```
jlizen: It's further complicated by `full` template which includes a ton of placeholders that mirror the individual templates.

Probably we should refactor to have it be a placeholder about whether to generate the scaffolding template, and anyway generate workflows as individual files.

md/rfds/categories-and-alternatives.md
```diff
@@ -0,0 +1,685 @@
...
+cargo bp add embedded -F stm32f4 -F embassy
+```
+
+Users who want deep chipset curation install the specific sub-pack:
+
+```bash
+cargo bp add stm32 -F stm32f4 -F embassy -F defmt-rtt
+```
+
+Both approaches compose: `stm32-battery-pack` could internally declare
```
jlizen: This is a bit annoying to need to need to keep features in sync if passing them through, but I guess it isn't so bad and avoids a lot of ambiguity?

It's clean how you can compose with regular cargo dependencies :)

jlizen: The more awkward part is if you start to have templates.

Do you inline any parent pack base functionality, and corresponding placeholders (`embassy` vs `rtic` setup, etc)?

I guess this is back to includes, and also not a new problem, so we don't need to deal with it yet.

It is already a small problem for the `logging-battery-pack` (and to lesser extent `error-battery-pack`), though neither are thick enough templates that it hurts to inline. Once we get to the second equivalent for `backend-service-battery-pack` it will hurt more though.

md/rfds/categories-and-alternatives.md
nikomatsakis: Yes, I was leaving that out for now, but it'd be easy enough to include

md/rfds/categories-and-alternatives.md
nikomatsakis: Whoops

md/rfds/categories-and-alternatives.md
```diff
@@ -0,0 +1,685 @@
...
+Features (and templates) that have no `category` annotation continue to work
+exactly as today — they appear as additive checkboxes in the picker.
+
+### Nesting: sub-categories
+
+For deeply structured domains like embedded (vendor → chip family → specific
+chip), categories can be nested with dot notation:
+
+```toml
+[package.metadata.battery-pack.categories.hal]
```
JayanAXHF: So if i'm mentioning this in a `[placeholder.whatever]`, i'd put this as `options = ["a.b"]` here right? What happens if between `A` and `B`, `A` is picked, but there is a prompt configured for `B`'s subcategories? the control flow would become more convolutied and harder to debug then

md/rfds/categories-and-alternatives.md
```diff
@@ -0,0 +1,685 @@
...
+  collapsible.
+- The header line says "(pick at most one)" for `at-most-one` categories.
+- **Validation on Enter**: if an `at-most-one` category has more than one
+  selection, the picker shows an inline error and refuses to confirm.
+
+### Non-interactive (CLI flags)
+
+```bash
+# Pick a specific HAL and concurrency framework:
+cargo bp add embedded -F stm32f4 -F embassy
```
JayanAXHF: for subcategories is it `a.b` here still?

md/rfds/categories-and-alternatives.md
JayanAXHF: +1 for this. How do templates interact with this? do they only get triggered when a specific choice path is taken?

JayanAXHF (commented):
Wouldn't it be better if we had a `battery-pack-rs/documents` or `rfd` repo?

md/rfds/categories-and-alternatives.md
nikomatsakis: I removed these subcategory things, they were overkill I think.

md/rfds/categories-and-alternatives.md
nikomatsakis: I removed this extension functionality.

md/rfds/categories-and-alternatives.md
JayanAXHF: That seems like a good move

md/rfds/categories-and-alternatives.md
nikomatsakis: I removed this

md/rfds/categories-and-alternatives.md
nikomatsakis: Fixed

md/rfds/categories-and-alternatives/impl-plan.md
JayanAXHF, line 1: How do templates interact with the alternatives?

JayanAXHF:
I'm strongly going to advocate for separating the two PRs; i think this has a value to be merged before the implementation is added

nikomatsakis:
> I'm strongly going to advocate for separating the two PRs; i think this has a value to be merged before the implementation is added

That's fine, happy to separate.

nikomatsakis:
Closing in favor of https://github.com/battery-pack-rs/battery-pack/pull/162

PR #160, opened 2026-07-08
feat(bp-managed): resolve managed deps in target-gated tables

### LINKED ISSUE #139: bp-managed not resolved in `[target.*.dependencies]`
`resolve_bp_managed_content` only scans `[dependencies]`, `[dev-dependencies]`, and `[build-dependencies]`. Deps under `[target.'cfg(...)'.dependencies]` keep their literal `bp-managed = true` (invalid Cargo). Target-gated deps (e.g. jemalloc, which can't build on MSVC) therefore can't use `bp-managed` and must be pinned explicitly in templates.

Fix: have the resolver also walk `target.<cfg>.{dependencies,dev-dependencies,build-dependencies}`.

Might be fixed by #135.

gmemuriuki:
Extend `bp-managed = true` resolution to `[target.<cfg>.*]` mirrors, not just the top-level sections.

One `for_each_dep_table` traversal now drives both the validation and rewrite passes over every `[dependencies]`, [`dev-dependencies`], and `[build-dependencies]` table and their per-target copies, matching entries as `TableLike` so block and inline syntax resolve identically.

Harden resolution while here:
- Renamed deps (`foo = { package = "bar", bp-managed = true }`) resolve by their real crate name, not the table key.
- A non-`true` marker (`bp-managed = false`, a string, ...) is now an error instead of shipping verbatim into the generated project.
- `workspace` joins `version` as a key that conflicts with `bp-managed`.

Model the marker as a `BpManaged` enum whose `Enabled` variant carries the entry's table view, so the rewrite pass never re-derives it.

Add regression tests for each case; convert the test TOML fixtures to `indoc!`.
Update the manifest spec and template-authoring docs.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

closes #139


JayanAXHF:
This failure is really weird; i hope it compiles on your machine?

JayanAXHF:
@gme-muriuki i've investigated the issue; this looks like an issue with the dependencies in either backend service bp or somewhere in the callstack of the test. i don't think your PR has affected anything

JayanAXHF:
I've confirmed that this PR is not at fault. Will be opening a related issue for the bug

jlizen:
```

    Checking tower-http v0.6.11
error[E0425]: cannot find type `PollTimeHistogram` in this scope
   --> /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-metrics-0.5.0/src/runtime.rs:524:34
    |
524 |         pub poll_time_histogram: PollTimeHistogram,
    |                                  ^^^^^^^^^^^^^^^^^ not found in this scope
    |
note: found an item that was configured out
   --> /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-metric
```


This is fixed upstream: https://github.com/tokio-rs/tokio-metrics/pull/131
Pending release

For now the workaround would be to pass the `--tokio-unstable` flag into CI

jlizen:
CI should be fixed by https://crates.io/crates/tokio-metrics/0.5.1, triggered a re-run

jlizen:
Looks like we might need to bump the `Cargo.lock`, actually, if anybody has time.

JayanAXHF:
Consider rebasing over the lock changes

gmemuriuki:
Thanks @JayanAXHF && @jlizen

md/creating.md
```diff
@@ -67,6 +67,24 @@ This reads your `Cargo.toml`, `README.md`, and `docs.handlebars.md`, then render
 ```handlebars
 \{{readme}}
 
+# Managed version with optional:
```
jlizen: This belongs in the `Managed  dependencies`  section instead of here.

jlizen (changes requested):

md/creating.md
gmemuriuki: It's already there actually, so I removed it.

jlizen (approved):

PR #175, opened 2026-07-14
fix(ci): fix mdbook workflow to support battery-pack preprocessor

nikomatsakis:
The preprocessor is invoked via `cargo mdbook-battery-pack` (a cargo alias for `cargo run -p mdbook-battery-pack --release`), which requires the Rust toolchain on PATH. Fix the rustup install invocation and ensure cargo/bin is in GITHUB_PATH so mdbook can find cargo when it shells out to the preprocessor.


nikomatsakis:
(I'm unconvinced this change makes sense)

PR #181, opened 2026-07-18
refactor(tui): modularize tui.rs into components and utils

### LINKED ISSUE #57: (refactor) Split up tui.rs into submodules
Currently, `tui.rs` is a 3800-line monolith that is very hard to work with and undestand. So we should 
- [ ] Move the tests into their own file/directory
- [ ] Move the `App` struct and adjacent code into the new module's root and 
- [ ] Move the components into their own module, or club small related ones into their own module
- [ ] Move the utility function into their own module

Stems from #48

shamnad-sherief:
Breaks down the large `tui.rs` monolith by introducing a `tui/` directory. Rendering logic is now in `components.rs` and generic helpers in `utils.rs`, leaving `mod.rs` to handle core App state and routing.

closes #57


jlizen (approved):
Thanks!

PR #187, opened 2026-08-06
perf(bphelper-cli): batch rustfmt into a single process per render

### LINKED ISSUE #184: [chore] Batch rustfmt calls during template rendering
`cargo bp show` currently starts one `rustfmt` process for every generated Rust file. For the
backend-service template, this means 11 formatter processes:

```bash
./target/release/cargo-bp bp --non-interactive show backend-service \
  --path battery-packs/backend-service-battery-pack \
  -t service -d service_name=demo -d description=demo
```

A release flame graph attributes 69% of baseline samples to `rustfmt`, mostly repeated process
startup and dynamic linking. Across 20 runs, the command averaged 216.78 ms.

A prototype that formats all unique Rust paths in one `rustfmt` invocation reduces the process
count from 11 to 1 and averages 82.43 ms, a 61.97% reduction. It preserves caller-side rustfmt
configuration, duplicate-path behavior, and all-or-nothing fallback on failure.

Would the project accept a focused PR with this batching change, regression tests, and the
before/after flame graphs?

Abusalah0:
## Summary

- Batch rustfmt formatting of rendered .rs files into a single process per render, instead of spawning one per file
- Preserve stdin-equivalent behavior by using --config skip_children=true (avoid resolving mod children from temp files) and --config-path (rediscover the caller's rustfmt.toml/.rustfmt.toml the same way stdin formatting did)
- Falls back to the previous per-file pipe approach for duplicate rendered paths (which can't share one temp file); still no-ops entirely on any rustfmt/temp-file failure so generation never hard-depends on rustfmt being installed

## Tests

- [x] cargo test -p bphelper-cli — new unit tests cover multi-file batching, duplicate paths, missing-rustfmt files, failure/read-failure rollback, and unsafe-path rejection
- [x] cargo test --all --workspace passes
- [x] cargo clippy --all-targets and cargo fmt --check clean
- [x] Manually verified against real rustfmt: skip_children=true prevents mod-resolution errors on temp files, and --config-path correctly picks up project rustfmt.toml/.rustfmt.toml with the same precedence as stdin formatting
- [x] Benchmarked: 30 files formatted individually (0.35s) vs. batched (0.014s), ~25x faster

Closes #184

### DIFF
--- a/src/battery-pack/bphelper-cli/src/template_engine.rs
+++ b/src/battery-pack/bphelper-cli/src/template_engine.rs
@@ -313,38 +313,148 @@ fn render(
 
     // Normalize rendered Rust through rustfmt so generated output is formatting-stable regardless
     // of template whitespace. No-ops if rustfmt is missing or rejects the input.
-    for file in files.iter_mut().filter(|f| f.path.ends_with(".rs")) {
-        file.content = rustfmt_rust(std::mem::take(&mut file.content));
-    }
+    rustfmt_rust_files(&mut files);
 
     Ok(files)
 }
 
-/// Formats Rust source with rustfmt (edition 2024) over stdin/stdout, with no temporary files so
-/// it works for previews too. Returns the input unchanged if rustfmt is unavailable or fails, so
-/// generation never hard-depends on a rustfmt install.
-fn rustfmt_rust(source: String) -> String {
+/// Formats rendered Rust files with rustfmt (edition 2024). Returns all files unchanged if rustfmt
+/// is unavailable or fails, so generation never hard-depends on a rustfmt install.
+fn rustfmt_rust_files(files: &mut [RenderedFile]) {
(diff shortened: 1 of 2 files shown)


src/battery-pack/bphelper-cli/src/template_engine.rs
```diff
@@ -313,50 +313,161 @@ fn render(
...
+        .enumerate()
+        .filter_map(|(idx, file)| file.path.ends_with(".rs").then_some(idx))
+        .collect();
+    if rust_files.is_empty() {
+        return;
+    }
+
+    // Preserve independent contents for duplicate rendered paths, which cannot share one temp file.
+    let mut rendered_paths = HashSet::with_capacity(rust_files.len());
+    if rust_files
```
JayanAXHF, line 344: If i understand the fallback mechanism correctly,  fallback triggers per-file piping for every Rust file in the render if any path collides. right?

